PR #2, opened 2018-10-03
readme and some basic file

bmnguyen:
I want to comfirm that his is how I can request files from the origin in my base repo to be pull into the share class room repos.


bmnguyen:
adding mailroom.py as another exercise

PR #21, opened 2018-10-06
Session02 grid_printer exercise

motians:


motians:
Added the grid_printer.py exercise with three function. One function for each part of the exercise.

students/shawn_degraw/session01/string-2.py
```diff
@@ -37,3 +37,28 @@ def end_other(a, b):
...
   return False
 
+#cat_dog
+def cat_dog(str):
+  catcount = 0
+  dogcount = 0
+  for i in range(len(str)-2):
+    comparestr = "cat"
+    nextstr = str[i:i+3]
+    if nextstr == comparestr:
```
hosungs, line 47: You might just have used `if str.startswith('cat', i):` without lines 45 & 46. You may refer to web pages like https://www.geeksforgeeks.org/python-string-startswith/ for that.

hosungs (approved):
Looks good! Thanks for your diligent work!

PR #22, opened 2018-10-10
Clean up directories within brandon_nguyen

bmnguyen:
I had to recreate the clone on my laptop.  I also learned how to remove a file and move file the correct way within git.


hosungs:
Thanks Brian! Looks good and I'm glad you found out various git techniques!

PR #23, opened 2018-10-11
Session 2 assignments pull request

funnyletter:


students/rachel_s/session01/pp_strings2.py
```diff
@@ -0,0 +1,69 @@
...
+        newstring = newstring + i + i
+    return newstring
+
+
+# Return the number of times that the string "hi" appears anywhere in the given 
+# string.
+def count_hi(str):
+    count = 0
+    for i in range(len(str) - 1):
+        if(str[i:i+2] == "hi"):
```
hosungs, line 15: It's really minor, but parenthesizing the entire if condition in python is not needed and actually not desired.

students/rachel_s/session02/grid_print.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def print_grid(n):
+    """Prints a grid of width and height n that is 2 cells high and wide."""
+    width = n // 2
+    # Print top border
+    print("+", end=" ")
+    for i in range(width):
+        print("-", end=" ")
+    print("+", end=" ")
```
hosungs, line 22: These 3 lines (lien 20-22) are repeated quite a few times in this function, so it would have been better if you created a helper function doing this and just call that function multiple times.

hosungs: Forgot to mention that there could be other such repeated code (possibly with slight differences that can be parameterized in functions) that could be refactored into subroutines (helper functions).

students/rachel_s/session02/grid_print.py
```diff
@@ -0,0 +1,101 @@
...
+        print("-", end=" ")
+    print("+")
+
+
+def print_grid2(grid_size, cell_size):
+    """Print a grid of a specified length and cell size
+
+    Keyword arguments:
+    grid_size: indicates the number of cells in the grid
+    cell_size: indicates the size of the cell"""
```
hosungs, line 67: Another really minor thing: It'd be better to clearly show that the docstring comments ends here by placing the closing `"""` in the next line at its beginning. Not sure if it's a requirement in docstring formatting, but I just found myself puzzled a bit about where the docstring ends (got so used to looking at the line beginnings only).

students/rachel_s/session02/grid_print.py
```diff
@@ -0,0 +1,101 @@
...
+    """Prints the internal part of a grid row, the bit with just vertical lines
+
+    Keyword arguments:
+    grid_size indicates the number of cells in the grid
+    cell_size indicates the size of the cell"""
+    for a in range(grid_size):
+        print("|", end=" ")
+        for b in range(cell_size):
+            print(" ", end=" ")
+    print("|")
```
hosungs, line 101: Well, you did the desired refactoring in this final part beautifully, so I guess you can ignore my comments above regarding refactoring... Great job!

hosungs (approved):
Excellent! You've done a lot of practices and your coding style and organization look really good. Great work!

students/rachel_s/session02/series.py
```diff
@@ -0,0 +1,54 @@
...
+    """Return the nth value in the Lucas numbers series starting with 0 index"""
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n-1) + lucas(n-2)
+
+
+def sum_series(n, n0=0, n1=1):
```
Saravji, line 19: It is great to see the effort you put in in going thru all the warm-ups!
So, at this point something to think about: fibonacci, lucas and sum_series are series summation functions. sum_series is a generalized form, while fibonacci and lucas are special cases with certain initial values.
Question: is it necessary to maintain a code base with series summation logic in three places?

hosungs: I think actually the intention is not to rewrite `fibonacci()` and `lucas()` with `sum_series()`, but to derive `sum_series()` from the two similar `fibonacci()` and `lucas()`, so I think there's no need to rewrite. Yes, we saw another student's submission that rewrote `fibonacci()` and `lucas()`, andI thought that's OK as well (but not necessary for the sake of this assignment).

PR #24, opened 2018-10-11
fizz_buzz.py and series.py complete

motians:


students/shawn_degraw/session02/series.py
```diff
@@ -0,0 +1,72 @@
...
+        else:
+            return sum_series(n-1) + sum_series(n-2)
+    """ compute the nth Lucas number """
+    if n0 == 2 and n1 == 1:
+        if n == 0:
+            return 2
+        elif n == 1:
+            return 1
+        else:
+            return sum_series(n-1, 2, 1) + sum_series(n-2, 2, 1)
```
hosungs, line 47: Hmm... I'm not sure if you need to distinguish the Fibonacci code and the Lucas code with hard-coded constant values, because in either case, you are returning `n0` if `n==0` and returning `n1` if `n==1`. And you are passing `n0` and `n1` to your recursive function calls anyway. So I think you can combine the two similar code blocks into just one and reduce the repetition.

hosungs (commented):
This PR was merged without a proper review (which we are struggling to define as appropriate, by the way), so you'll need to submit another PR if you'd like us to review more.

@Saravji -- Let's not merge PRs until the assignment due is over, so that all of us can iterate over open/pending PRs, not having to create more PRs. And please do some code review and provide feedback (just lgtm is fine as well) before merging.

students/shawn_degraw/session02/series.py
```diff
@@ -0,0 +1,72 @@
...
+
+def lucas(n):
+    """
+    compute the nth Lucas number
+    
+    :param n: value of the nth lucas element
+    """
+    return sum_series(n, 2, 1)
+
+def sum_series(n, n0=0, n1=1):
```
Saravji, line 18: Your sum_series function only works for parameters n0 and n1 that are set equal to those of fibonacci or lucas series.
The idea is to have this function as a more generalized form of fibonacci and lucas and make it universally valid.

students/shawn_degraw/session02/fizz_buzz.py
```diff
@@ -0,0 +1,12 @@
+#fizz buzz function
+for i in range(1,101):
+    result3 = i % 3
```
Saravji, line 3: It is not necessary to define variables and assign the results of a simple math operation to use these results in if statements.

students/shawn_degraw/session02/series.py
```diff
@@ -0,0 +1,72 @@
...
+    """ compute the nth Lucas number """
+    if n0 == 2 and n1 == 1:
+        if n == 0:
+            return 2
+        elif n == 1:
+            return 1
+        else:
+            return sum_series(n-1, 2, 1) + sum_series(n-2, 2, 1)
+    """ Other series: not implemented yet """
+    return None
```
hosungs, line 49: So, this was left not done? I was expecting this to be completed, leading to the generalization of the above two cases. Geez, those comments in the template did seem to have confused many students...

PR #25, opened 2018-10-11
week 2 submission 

aminavi:


students/aminavi/Lightning.py
```diff
@@ -0,0 +1,34 @@
...
+Guo Ying
+Jakubiak    Jason
+Meese   Tim
+Minavi  Alexander
+Morrison    Isaac
+Nalla   Arun K
+Nguyen  Brandon Q
+Schirra Rachel Elizabeth
+Song    Yushu
+Tetteh  Laud N
```
hosungs, line 18: Oops, this is not good, to show all names. Please remove this file from your directory. You already have this file in the `examples` directory.

students/aminavi/session02/grid3.py
```diff
@@ -0,0 +1,7 @@
+def grid(x,y):
+    for i in range(y*2 + 3):
+        if i%(y + 1)==0:
+            print("+" + "-"*x + "+" + "-"*x + "+")
+        else:
+            print("|" + " "*x + "|" + " "*x + "|")
+grid(1,1)
```
hosungs, line 7: I'm not sure if you understood this problem correctly? Do you get the same result for like `grid(5,11)`, as shown in the problem description? I don't think so, and I think this code will need to be corrected accordingly. This part can be nontrivial, so please try more.

hosungs (commented):
@Saravji -- Again, we shouldn't merge assignment PRs until the due is over. Let's keep all PRs open until the due is over.

And let's make sure that we require some note in any PR if students want us to review their code in advance, and if so, we do provide code review feedback like I'm doing. Otherwise, a PR will be just considered as a required submission as stated, and won't get code reviews, other than your final grading results. Hope this helps, and let me know if you have any comments. Thanks.

PR #26, opened 2018-10-11
Added folder session02 and Fizzbuzz.py

davidarasim:
Dave Arasim's Fizz Buzz assignment for Lesson 2

### DIFF
--- a/students/dave_arasim/session02/fizzbuzz.py
+++ b/students/dave_arasim/session02/fizzbuzz.py
@@ -1,0 +1,12 @@
+#Fizz Buzz Exercise for Lesson 2
+
+for x in range(100):
+    x += 1 #changes range 0-99 to 1-100
+    if not(x%3) and not(x%5):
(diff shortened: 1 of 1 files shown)


students/dave_arasim/session02/fizzbuzz.py
```diff
@@ -0,0 +1,12 @@
+#Fizz Buzz Exercise for Lesson 2
+
+for x in range(100):
+    x += 1 #changes range 0-99 to 1-100
```
hosungs, line 4: Hmm... I learned something new from this (I left an incorrect comment myself). I guess this works (I was mistaken earlier), but this is not very desired.

To get 1, 2, 3, ..., 100, this line should be removed and the `range(100)` should be changed to `range(1, 101)`. Search the web for `python range` and study how it's used.

hosungs (commented):
@Saravji -- Again, please let's not merge PRs until the due is over.

PR #27, opened 2018-10-11
submit fizzBuzz needing PR link

bmnguyen:
I just need the link so I can submit the exercise fizzBuzz.


students/brandon_nguyen/session02/fizzBuzz.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+#Fizz Buzz problem - Week 2 Excercise 1
+#Student: Brandon Nguyen - Au2018
+for i in range(100):
+    i +=1
```
hosungs: Well, I'm modifying my previous comment, because I was wrong (mistaken earlier). I confirmed that this works, but this is not very good. To iterate over 1, 2, 3, ..., 100, your best bet would be to use `range(1,101)` in line 4, without this line 5. Search the web for `python range` and study how the `range()` function is used in Python.

hosungs (changes requested):
Not sure if you already added all files for this week's assignments, but in this PR only, I don't see all the files. I think this PR is best kept open until past the due. I'm still leaving my PR feedback, because I think I see something that's definitely incorrect. Also, please leave a note in your PR if you'd like us to review your code. Otherwise, we'll have to assume that it's just for the required submission. Hope this helps.

students/brandon_nguyen/session02/fizzBuzz.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+#Fizz Buzz problem - Week 2 Excercise 1
+#Student: Brandon Nguyen - Au2018
+for i in range(1,101):
+    if (i%3 ==0 and i%5 == 0):
```
hosungs, line 5: Minor: No need to surround if conditions with parentheses in Python (it's in fact discouraged). No need to change that here, just remember for your future coding.

hosungs (commented):
I assume you'll add more commits as you complete all the assignments, so leaving the PR open until the due is over as mentioned (well, that's the plan no matter what anyway...)

students/brandon_nguyen/session02/fizzBuzz.py
bmnguyen: thank you!

bmnguyen: Please review my series.py and provide feedback.  I was trying to find a proper way to check args but unable to find a "correct" way to check the parameters for type etc.

students/brandon_nguyen/session02/series.py
```diff
@@ -0,0 +1,90 @@
...
+    if (n0==0 and n1==1):
+    #do fibonacci
+        if (n==0 or n==1): 
+            return n
+        else:
+            return fibonacci(n-1) + fibonacci(n-2)
+    elif(n0==2 and n1==1):
+    #do lucas    
+        if (n==0 or n == 1):
+            return abs(n-2)
```
hosungs, line 47: Do you see that you are returning `n0` if `n==0` and returning `n1` if `n==1`? Not just here, but also on the Fibonacci part as well (lines 39-40). This shows that you really don't need to distinguish the two series (Fibonacci and Lucas) at all. In fact, the purpose of `sum_series` is to unify all such series in one simple function, without having to check the parameter values like this and distinguished computation. Hope this helps.

hosungs (commented):
Left a comment to your series.py as you requested.

students/brandon_nguyen/session02/series.py
bmnguyen: Thank you!  I have made my correction.

students/brandon_nguyen/session02/series.py
```diff
@@ -0,0 +1,102 @@
...
+        print("TODO - Other Series sum to do")
+
+    '''
+    #correction after reviewed by instructor
+    if n==0:
+        return n0
+    elif n==1:
+        return n1
+    elif n >=2:
+        return sum_series(n-1,n0=n0,n1=n1) + sum_series(n-2,n0=n0,n1=n1)
```
hosungs, line 60: Yes, this is it! You can safely remove the original code.

Note that there's no need to enter `sum_series(n-1,n0=n0,n1=n1)`. Just use `sum_series(n-1,n0,n1)`. Just a minor comment.

hosungs (commented):
Good job with `sum_series()`!

students/brandon_nguyen/session02/series.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    compute the nth value for fibonacci or lucas serie. 
+    :param optional or default to n0=0, n1=1 determine the first two values in the series. 
+
+    if n0 == 0 and n1 == 1 do Fibbonacci 
+    if n0 == 2 and n1 == 1 do lucas
+    elif do other 
+
+    """
+    #ask the instructor which is the best way to check args.
```
Saravji, line 36: Brandon,
you have it the wrong way around here: Instead of using the fibonacci function or lucas function for a specific subset of sum_series, approach it as sum_series is a more generalized version of fibonacci and lucas. Or put in another way, fibonacci is a sum_series with certain n0 and n1 values, while lucas is a sum_series with certain (but different) values for n0 and n1.
I let you think about where you can take it but encourage you to review your solution once again. More for the understanding then for function.
No need to change it, it works as intended.

Saravji (approved):

PR #28, opened 2018-10-11
session02 commit added: grid, fizzbuz, series, break_me

jjakubiak:


students/jjakub/lighting.py
```diff
@@ -0,0 +1,34 @@
...
+Jakubiak    Jason
+Meese   Tim
+Minavi  Alexander
+Morrison    Isaac
+Nalla   Arun K
+Nguyen  Brandon Q
+Schirra Rachel Elizabeth
+Song    Yushu
+Tetteh  Laud N
+'''
```
hosungs: Oops, it's not good to show all the real names here. Please consider remove this file from your directory and push that removal change. This file is already available in the `examples` directory.

students/jjakub/session02/grid01.py
```diff
@@ -0,0 +1,9 @@
+
+plus = "+----+----+"
+pipe = "|    |    |"
+
+for i in range(11):
+    if i%5 == 0:
+        print(plus)
+    else:
+        print(pipe)
```
hosungs, line 9: Smart!

students/jjakub/session02/grid03.py
```diff
@@ -0,0 +1,12 @@
+def grid_fcnt(x,y):
+
+    plus = "+" + "-"*x + "+" + "-"*x + "+"
+    pipe = "|" + " "*x + "|" + " "*x + "|"
```
hosungs: I think you misunderstood the problem. There should be `y` many segments each of which should be `x` many characters. You still have fixed 2 segments of `x` many characters. Same goes for the rows as well, but I think your `for` loop below still prints just 2 segments, not `y` many segments.

hosungs (changes requested):
Please see my comments. Will leave this PR open until the assignment due is over, so that any further commits you push will pile up here.

students/jjakub/session02/grid03.py
```diff
@@ -0,0 +1,12 @@
+def print_grid2(x,y):
+
+    plus = (("+" + "-"*y) * x) + "+"
+    pipe = (("|" + " "*y) * x) + "|"
+
+    for row in range((y * x) + x + 1):
+        if row%(y + 1) == 0:
+            print(plus)
+        else:
+            print(pipe)
```
hosungs, line 10: Interesting, seemingly smart logic, and I think it's going to work (purely from reading the code), but I'll let TA do the more in-depth grading.

students/jjakub/session02/series.py
```diff
@@ -0,0 +1,59 @@
...
+    elif n==1:
+        return 1
+    else:
+        return lucas(n-1)+lucas(n-2)
+
+luc_result = lucas(8) # Input
+print(luc_result)
+
+# Sum Series
+def sum_series(n, val01=0, val02=1):
```
Saravji, line 26: Try to use more descriptive variable names. It is simple right now, but once the logic gets more complex, you want to avoid confusion and possibly wrong assumptions because of variable names.

students/jjakub/session02/grid03.py
Saravji: Indeed, a very clever and elegant solution. (and correct) Also, it is relatively good readable.

Saravji (approved):

PR #29, opened 2018-10-13
Session02 Assignments: GridPrinter, FizzBuzz, Fibonacci Series

laudtetteh:


students/laud/session02/lightning.py
```diff
@@ -0,0 +1,35 @@
...
+Guo Ying
+Jakubiak    Jason
+Meese   Tim
+Minavi  Alexander
+Morrison    Isaac
+Nalla   Arun K
+Nguyen  Brandon Q
+Schirra Rachel Elizabeth
+Song    Yushu
+Tetteh  Laud N
```
hosungs: I should have mentioned specifically not to copy/create this file in your directory... It's not good to show all these names in public, so please remove this file from your directory. You can get the redacted source file from `examples` directory already.

hosungs (commented):
Seeing only fizzbuzz.py included, I'll expect that you'll push more commits for the other 2 assignments as well, so I'll leave this open. I suppose you might have meant to create different PRs for different assignments, but we won't be able to quickly review/merge individual PRs like that, so we are using only 1 PR for all 3 assignments that are due next Tuesday. You might want to modify the PR title to reflect that. Thanks.

students/laud/session02/lightning.py
laudtetteh: @hosungs Thanks for pointing that out. I remove it.

laudtetteh:
@hosungs I didn't get a chance to get a good grasp of the Fibonacci/Lucas assignment, so you'll notice my attempt was pretty dismal. I'm hoping to read around some more and catch up soon!

hosungs:
> @hosungs I didn't get a chance to get a good grasp of the Fibonacci/Lucas assignment, so you'll notice my attempt was pretty dismal. I'm hoping to read around some more and catch up soon!

No worries! Since the due is over, TA will go over the submissions and provide feedback. Thanks!

students/laud/session02/Series.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+# Fibonacci Series
+def fibonacci(n):
+    a, b = 0, 1
+    series =  []
+    while (b < 20):
```
Saravji, line 6: please review the concept of recursion. 
This should solve the difficulty that you had here and that you trying to solve with this while construct.

Saravji: I think, you should attempt to still solve this assignment. It is outside the grading, but the learning that you can take from it would be beneficial. You have an interesting and promising start, so keep going and try to solve it!

PR #30, opened 2018-10-15
PR for assignments 2

yushusongmsazure:
This PR includes solutions for session2:
1. fizz_buzz.py
2. series.py
3. print_grid.py
4. print_grid1.py
5. print_grid2.py

Please ignore break_me.py


students/yushus/session01/break_me.py
```diff
@@ -0,0 +1,31 @@
+import sys
+
+def yushus_Exception_NameError():
+	try:
+		raise NameError	
+	except NameError:
+		print("Catched NameError")
```
hosungs: I saw your message asking to ignore `break_me.py`, but just wanted to let you know that a model solution `break_me.py` is posted at [here](https://github.com/UWPCE-PythonCert-ClassRepos/Au2018-Py210B/blob/master/solutions/Lesson01/break_me.py), so you might want to take a look. The purpose of this activity is not to use the `raise` statements to raise those exceptions, but to give some other statements that will raise those exceptions. The solution `break_me.py` does show those statements, so please check it out.

Also it looks like your indentation is more than 4 spaces, but it also looks like it's fixed in other/later source files, so just mentioning here.

students/yushus/session02/print_grid/print_grid.py
```diff
@@ -0,0 +1,26 @@
+# control the height or how many rows to print
+for l in range(2):
+    # print the first + sign
+    print('+', end='')
+
+    # control the length or how many columns to print
+    for i in range(2):
+        # one iteration looks like: ----+
+        print('-'*4,end='')
```
hosungs: Minor: I think there's a blank space for each `-` character printed, but I'd say it's not a big deal. You may want to compare your outputs with the exact printouts from the example outputs.

students/yushus/session02/series.py
```diff
@@ -0,0 +1,67 @@
...
+    if n0 == 0 and n1 == 1, the result is the Fibbonacci series
+
+    if n0 == 2 and n1 == 1, the result is the Lucas series
+    """
+    if n0 == 0 and n1 == 1:
+        return fibonacci(n)
+    elif n0 == 2 and n1 == 1:
+        return lucas(n)
+    else:
+        raise "Not the right format!!!"
```
hosungs: This is not correct. The `sum_series()` function should be applicable for any `n0` and `n1` values, not just for Fibonacci or Lucas series. Basically your `sum_series()` function is expected to call itself recursively so that it'll calculate any sum series, including Fibonacci and Lucas series. The key point is that your `sum_series()` should return `n0` if `n==0`, or return `n1` if `n==1`, or return the corresponding recursive result. Please fix.

hosungs (changes requested):
Comments left for you to follow up... Thanks!

students/yushus/session02/print_grid/print_grid.py
yushusongmsazure: I see. Included the space after each -

students/yushus/session02/series.py
yushusongmsazure: Thank you so much Hosung. I see my mistake now :-). Changed.

students/yushus/session02/series.py
```diff
@@ -36,12 +17,21 @@ def sum_series(n, n0=0, n1=1):
...
-        return lucas(n)
+        if n == 0:
+            return 2
+        elif n == 1:
+            return 1
+        else:
+            return (sum_series(n-1, n0, n1) + sum_series(n-2, n0, n1))
     else:
-        raise "Not the right format!!!"
+        raise "Function not defined"
```
hosungs: Still not correct... `sum_series()` is defined for any input values of `n0` and `n1`. You really don't need to check if `n0==0 and n1==1` or `n0==2 and n1==1`. Just forget about that checking, and return `n0` if `n==0`, `n1` if `n==1`, and the recursive result. Note that you have repeated code that's not needed and should be avoided.

hosungs (commented):
Just responding to the auto-notification...

students/yushus/session01/break_me.py
yushusongmsazure: Thank you!

students/yushus/session02/series.py
```diff
@@ -0,0 +1,48 @@
+def fibonacci(n):
+    """ compute the nth Fibonacci number """
+    return sum_series(n, 0, 1)
+
+def lucas(n):
+    """ compute the nth Lucas number """
+    return sum_series(n, 2, 1)
+
```
Saravji, line 8: I like that you took it to this abstraction level and are using your generalized sum_series function to compute fibonacci and lucas, set with the respective initial elements.
Well done!

PR #31, opened 2018-10-16
 Lesson02 activity

nallarun10:
Finished lesson02 activities.
Had some issues with where to add 'return' and 'print' values in functions?
Thanks
Arun


students/arun_nalla/session02/grids_V2.py
```diff
@@ -0,0 +1,47 @@
...
+        for a in range(n+2):
+            if a==0 or a==(n+1)//2 or a==n+1:
+                print (y)
+            elif a!=0 and a!= (n+1)//2:
+                print (x)
+        
+def grid2(n1,n2):
+    x = ('| '+'  '*(n2))*n1+'|'
+    y = ('+ '+'- '*(n2))*n1+'+'
+    v =[y if a==0 else x for a in range(n2+1)]
```
hosungs, line 23: Great!

students/arun_nalla/session02/series.py
```diff
@@ -0,0 +1,78 @@
+﻿#!use/bin/env python
+def fibo(n):
+    '''Write a alternative code to get the nth value of the fibonacci series using type (list)
+    0th and 1st position should return 0 and 1 and indexing nth position should
+    return SUM of previous two numbers'''
+    if n ==0: return 0
+    elif n ==1: return 1
+    my_list = [0,1]
+    for x in range(n+1):
+        if x>=2:
```
hosungs: Instead of these 2 lines (9-10), you could have just `for x in range(2, n+1):`. You may want to look up the documentation for `range()` and get familiar with its various usages.

students/arun_nalla/session02/series.py
```diff
@@ -0,0 +1,78 @@
...
+            y = my_list[-2] + my_list[-1]
```
hosungs: Wow, great usage of negative indices!

students/arun_nalla/session02/series.py
```diff
@@ -0,0 +1,78 @@
...
+    two keyword agruments decides the return value.  The default values of keyword arguments be 1 and 0.
+    The default keyword arguments results in fibonacci series and return nth position of based on positional arguments
+    if keyword arguments were 2 and 1, the function return Lucas series
+    for any other value entered should results sum of all the enters'''
+    if a ==0 and b==1:
+        print (fibo(n))
+    elif a==2:
+        print (lucas(n))
+    else:
+        print (n+a+b)
```
hosungs: You did great on Fibonacci and Lucas, but this `sum_series` needs to be more generalized. You are not supposed to distinguish your `sum_series` computation into `fibo` or `lucas` by checking the parameter values, but you are to generalize all such sum series computation (that includes Fibonacci and Lucas) by utilizing the general rule (in your parameter naming, `sum_series` should return `a` if `n==0`, return `b` if `n==1`, and return the sum_series result, but your initial `my_list` should start as `[a,b]`. Hope this makes sense.

One minor thing: You don't print the result in the function, but just return it.

hosungs (changes requested):
Left some comments.

students/arun_nalla/session02/series.py
```diff
@@ -0,0 +1,97 @@
...
+    if a==2 and b ==1:
+        list_sum_series = [a,b]
+        if n == 0: return 2
+        elif n == 1: return 1
+        for x in range(2, n + 1):
+            y = list_sum_series[-2] + list_sum_series[-1]
+            list_sum_series.append(y)
+        return list_sum_series[-1]
+    else:
+        return n+a+b
```
hosungs: I'm reviewing again, as you pushed new commits, but not sure if this is the latest code and this will handle the `sum_series` computation as desired? Again, your `sum_series` is not to distinguish between Fibonacci and Lucas, by checking the `a` and the `b` parameters. Instead, your `sum_series` should generalize all the possible sum series by unifying the necessary logic into just 1 block of code. No outside `if a==0 and b==1:` kind of distinction. Also, your current code doesn't properly compute sum series that's neither Fibonacci nor Lucas, so it still needs fixing.

hosungs (commented):
Left some comment after the latest pushed commit...

PR #32, opened 2018-10-16
correction grid 3 r1

aminavi:


students/aminavi/session02/grid3.py
```diff
@@ -1,7 +1,16 @@
...
+    for k in range(x):
+        for j in range(y+1):
+            if j == 0:
+                print("+", end="")
+                for i in range(x):
+                    print(y*"-", end='+')
+            else:
+                print("|", end="")
+                for i in range(x):
+                    print(y*" " , end='|')
```
hosungs, line 11: Good work, but if you think about it, then you might be able to unify the two blocks of code in lines 5-7 and in lines 9-11. No need to do that for this assignment. We'll learn how to do more refactoring as we go along.

PR #33, opened 2018-10-17
homework for session 2

guoguoying2013:
I pushed some homework to my origin master yesterday. Today I made some correction and pushed to origin master again. I can see in this PR, it both shows the old version and current version. Please see the current version as my submission of homework.

Thank you and regards,

Ying

### DIFF
--- a/students/YingGuo/session02/GridPrinter.py
+++ b/students/YingGuo/session02/GridPrinter.py
@@ -1,0 +1,87 @@
+
+def print_function1():
+    """homework2 #1, print a grid as shown in homework"""
+    print("+ - - - - + - - - - +")
+    for i in range(4):
+        print("|         |         |")
+    print("+ - - - - + - - - - +")
+    for i in range(4):
+        print("|         |         |")
+    print("+ - - - - + - - - - +")
+    return
+
+
+
+def print_grid(n):
+    """make the print grid function more general.
+    This is a function print_grid(n) that takes one integer argument and prints a grid just like before,
+    BUT the size of the grid is given by the argument,which is n"""
+    for x in range(n+2):
+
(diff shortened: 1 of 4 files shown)


students/YingGuo/session02/GridPrinter.py
```diff
@@ -0,0 +1,87 @@
...
+            for i in range(n+2):
+                if i == 0:
+                    print("|", end=" ")
+                elif i == (n+2)//2:
+                    print("|", end=" ")
+                elif i == n+1:
+                    print("|", end=" ")
+                else:
+                    print(" ", end=" ")
+            print()
```
hosungs, line 67: There's a lot of repeated code in this function, and it'd be desirable to avoid such repetition as much as possible. It looks like you achieved that in the next part, so I'd say it's OK here, but always remember the DRY (Don't Repeat Yourself) principle (though it might not make much sense now, hopefully it becomes more evident as we go along).

students/YingGuo/session02/series.py
```diff
@@ -0,0 +1,80 @@
...
+    list1 = list()
+
+    for i in range(n+1):
+        if i == 0:
+            list1.append(0)
+        elif i == 1:
+            list1.append(1)
+        else:
+            x = list1[i-2]+list1[i-1]
+            list1.append(x)
```
hosungs, line 17: This works, but it'd be better if you don't need to use a list of many numbers just to compute the final value. No need to fix, and the solutions that'll be posted are not most efficient anyway, so this is fine.

hosungs (commented):
Just left some minor comments. Good work overall!

Saravji (approved):

PR #34, opened 2018-10-17
session02 homework

ejackie:


students/ejackie/session02/fizzbuzz.py
```diff
@@ -0,0 +1,14 @@
+# numbers between 1 to 100 inclusive
```
Saravji, line 1: Please use a header in all your files. Should include Author, Date, Version and a brief description

students/ejackie/session02/series.py
```diff
@@ -0,0 +1,66 @@
...
+    """ compute the nth Lucas number """
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n-2) + lucas(n-1)
+
+
+def sum_series(n, n0=0, n1=1):
```
Saravji, line 28: This function was intended to be universally correct. Your implementation only works for fibonacci and lucas series.
E.g. it would not work for a series with n0=5 and n1=3
The idea here was to add one layer of abstraction.

Saravji (approved):

PR #35, opened 2018-10-17
Lesson 2 programs

davidarasim:


students/dave_arasim/session02/print_grid.py
```diff
@@ -0,0 +1,100 @@
...
+
+def print_grid_center2(grid_vert, grid_space, grid_len, square_ct): 
+    if grid_len > 0:
+        for grid_ct in range(grid_len):
+            print(grid_vert, end='')       #Left side (always printed)
+
+            for this_square in range(square_ct):
+                print(grid_space*grid_len, end='')
+                print(grid_vert, end='')
+            print()                        #New line
```
hosungs, line 84: Very well defined naming of variables/functions, well decomposed design (with lots of helper functions)! May look slightly over-designed for a simple problem like this, but I think it's better to approach this way. Good job, assuming code works as desired (which will be checked by TA)!

hosungs (approved):
Looks good! TA will review final code and merge then. Thanks for your hard work!

Saravji:
Good work!

PR #36, opened 2018-10-20
Session 03 assignemnt PR

funnyletter:


students/rachel_s/session03/list_lab.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+
+
+# Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches”.
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+og_fruits = fruits.copy()
```
hosungs, line 6: Minor: Wow, I never knew this and learned this from you on this occasion! I think most people would still recommend you to use `fruits[:]` instead of `fruits.copy()`, though. Just a little minor thing...

students/rachel_s/session03/list_lab.py
```diff
@@ -0,0 +1,102 @@
...
+
+# Display the list.
+print(fruits)
+
+# Ask the user for a number and display the number back to the user and the 
+# fruit corresponding to that number (on a 1-is-first basis). Remember that 
+# Python uses zero-based indexing, so you will need to correct.
+my_num = input("Pick a number between 1 and 6 > ")
+
+if int(my_num) in range(7):
```
hosungs, line 23: Minor: If user enters 0, this will pass, causing line 24 to crash (index out of bounds). I think you'd want to change the `range()` function params to avoid that.

students/rachel_s/session03/list_lab.py
```diff
@@ -0,0 +1,102 @@
...
+        print(i)
+print()
+
+
+# Display the list.
+fruits = og_fruits.copy()
+print(fruits)
+
+# Remove the last fruit from the list.
+del fruits[0]
```
hosungs, line 50: Isn't this the first fruit in the list?

students/rachel_s/session03/list_lab.py
```diff
@@ -0,0 +1,102 @@
...
+# For each “no”, delete that fruit from the list.
+# For any answer that is not “yes” or “no”, prompt the user to answer with one of those two values (a while loop is good here)
+# Display the list.
+
+fruits = og_fruits.copy()
+
+to_remove = []
+for i in fruits:
+    response = input("Do you like {}? ".format(i.lower())).lower()
+    while response not in ["yes", "no"]:
```
hosungs, line 80: Great!

students/rachel_s/session03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+            return item
+
+with open('donors.csv') as csv_file:
+    csv_reader = csv.reader(csv_file, delimiter=',')
+    for line in csv_reader:
+        if not line[0].lower() in donor_names:
+            donor_names.append(line[0].lower())
+            add_new_donor(donor_info, line[0].lower(), int(line[1]))
+        else:
+            add_donation(donor_info, line[0], int(line[1]))
```
hosungs, line 30: Wow, you already implemented the file reading feature, which is covered in the next week, not this week. Excellent!

hosungs (commented):
Excellent work! I haven't checked if all assignments are solved, but I'm pretty sure you have made almost to the completion quite early. Feel free to keep pushing any more commits and I'll let TA finish this review. Thanks for your hard work!

hosungs (approved):
Seeing that your week 4 commits are piling on this PR, we should close this ASAP. Great work!

PR #37, opened 2018-10-24
Session03, three labs and one assignment submitted on Oct 23rd

guoguoying2013:


students/YingGuo/session03/Slicing.py
```diff
@@ -0,0 +1,52 @@
+#hw3, slicing practice
+
+def sequence_1(sequence):
+    """first and last items exchanged"""
+    first = sequence[0]
+    last = sequence[-1]
+    rest = sequence[1:-1]
+    new_sequence = last + rest + first
```
hosungs, line 8: Does this work for any sequence, other than strings? I doubt this would work for a list (like `[1,2,3,4]`) for `sequence`. That's because `sequence[0]` will be an item (in the example, it'll be `1`, not `[1]`), same for `sequence[-1]` (in this example, it's `4`, not `[4]`). However, `rest` will be a list (`[2, 3]`) and Python doesn't allow `+` on a list object and a non-list object. Try calling this function with a tuple and see if it works. To fix that, you'll want to get `[1]` for `first` in the example, and think about how to achieve that.

students/YingGuo/session03/Slicing.py
```diff
@@ -0,0 +1,52 @@
...
+
+def sequence_2(sequence):
+    """every other item removed"""
+    sequence = sequence[::2]
+
+    return sequence
+
+def sequence_3(sequence):
+    """with the first 4 and the last 4 items removed, and then every other item in the remaining sequence"""
+    sequence = sequence[4:-5:2]
```
hosungs, line 20: Are you sure it's really `-5` here?

students/YingGuo/session03/Slicing.py
```diff
@@ -0,0 +1,52 @@
...
+    """revers elements with slicing"""
+    sequence = sequence[::-1]
+
+    return sequence
+
+def sequence_5(sequence):
+    """with the last third, then first third, then the middle third in the new order."""
+    sequence_unit = int(len(sequence)/3)
+    last_third = sequence[-sequence_unit:]
+    first_third = sequence[0:sequence_unit]
```
hosungs, line 34: Minor: `0` for the starting index can be omitted.

students/YingGuo/session03/list.py
```diff
@@ -0,0 +1,99 @@
...
+l = ["Kiwi"] + l
+print(l)
+
+#Add another fruit to the beginning of the list using insert() and display the list.
+l.insert(0,"Banana")
+print(l)
+
+#Display all the fruits that begin with “P”, using a for loop.
+for i in range(len(l)):
+    if l[i][0] is "P":
```
hosungs, line 32: Note that we should be careful with `is`, and this case doesn't seem appropriate for `is`. You want to check if the values of the two objects are the same, not if the ids of the two objects are the same (though it may work due to Python runtime's optimization for short strings like these).

students/YingGuo/session03/list.py
```diff
@@ -0,0 +1,99 @@
...
+        print(l[i])
+    else:
+        pass
+
+
+"""Series 2"""
+#Display the list.
+import copy
+l2 = copy.deepcopy(l)
```
hosungs, line 41: Hmm... It's great that you found this, but I'm not sure if this is an intended code for this activity... I think we want to exercise the plain old `seq[:]` to get a copy of `seq`. Yes, it's a shallow copy, but we are OK with a shallow copy here because all items are immutable (strings). We can talk about that later in class.

students/YingGuo/session03/list.py
```diff
@@ -0,0 +1,99 @@
...
+
+if user_input2 in l2:
+    l2.remove(user_input2)
+
+print(l2)
+
+
+"""Series 3"""
+#using the list from series 1
+import copy
```
hosungs, line 61: No need to import the same module again. Also, I should have mentioned earlier that `import` statements should be placed at the top of every source file. That's a good convention we should obey.

students/YingGuo/session03/list.py
```diff
@@ -0,0 +1,99 @@
...
+l3 = copy.deepcopy(l)
+"""Ask the user for input displaying a line like “Do you like apples?” for each fruit in the list (making the fruit all lowercase).
+For each “no”, delete that fruit from the list.
+For any answer that is not “yes” or “no”, prompt the user to answer with one of those two values (a while loop is good here)
+Display the list."""
+for x in l3:
```
hosungs, line 67: You are modifying `l3` inside the loop (`l3.remove(x)` is line 71), so it's best to iterate over its copy (`l3[:]`), not directly over that modified list (because the loop iteration sequence might be affected by the remove operations).

students/YingGuo/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
+#hw3, mailroom
+
+data = {"Tom":[100,200], "Tim":[3000], "Ken": [888,888,888], "Ted":[8000,5000,500], "Jon":[1000,6000]}
```
hosungs, line 3: Oh, you are already using a dict! :) I think that's fine, though it hasn't been covered in class yet (this week's topic).

students/YingGuo/session03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+			return data
+			break
+			
+		else:
+			amount2 = int(input("how much would you like to donat?"))
+			data[user_input.capitalize()] = list()
+			data[user_input.capitalize()].append(amount2)
+			print("Thank you {} for your donation {}!".format(user_input.capitalize(), amount2))
+			return data
+			break
```
hosungs, line 28: Notice that your lines 16-20 and lines 23-28 are almost identical, except you have to add line 24. Python offers a nice way to avoid this kind of code duplication using `.setdefault()` on a dict object. We'll see it together in this week's class. Just wanted you to notice this and wanted to let you know that we should always try to avoid duplicate code as much as possible (DRY--Don't Repeat Yourself).

hosungs (commented):
Took just a quick look and left some comments. I'll let TA follow up & finish (merge) PR. Good work and feel free to push more commits. Thanks!

hosungs (approved):
Not sure if my comments were addressed, but approving & merging this PR for the next assignment.

guoguoying2013:
So there is a second chance to improve homework and PR again? 
I though I can only PR once for each assignment.

hosungs:
You can still update your earlier code for improvement, and feel free to include those in your upcoming/future PR. I'll be happy to review those more, but TA's grading will be based on your Canvas submission. We need to merge/close PRs as early as possible, to avoid piling commits for the next week's assignment on this week's PR.

PR #38, opened 2018-10-24
PR for assignments 4

yushusongmsazure:


students/yushus/session03/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+    fruits.remove(fruits[-1])
+    print(fruits)
+
+    fruit_to_delete = input("Please choose a fruit to delete: ")
+    for f in fruits:
+        if f.lower() == fruit_to_delete.strip().lower():
+            fruits.remove(f)
+    print(fruits)
+
+    new_fruits = fruits + fruits
```
hosungs: Minor: Or `new_fruits = fruits * 2`:)

students/yushus/session03/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+    found = False
+    while(not found):
```
hosungs: Minor: No parentheses needed in Python's `while` and `if` conditions

students/yushus/session03/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+        fruit_to_delete = input("Please try again: ").strip()
+        for f in new_fruits:
+            if f.lower() == fruit_to_delete.lower():
+                new_fruits.remove(f)
```
hosungs: You are modifying the `new_fruits` list here, and yet you are iterating over the `new_fruits` list in line 51. Because the list may be modified (some items removed) in the loop, the loop's behavior might be unpredictable. Better approach is to iterate over a copy of `new_fruits` in line 51: `for f in new_fruits[:]:'

students/yushus/session03/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+                found = True
+    print(new_fruits)
+
+def series_3():
+    fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+    fruits = [f.lower() for f in fruits]
```
hosungs, line 61: Nice list comprehension!

students/yushus/session03/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+    return sum(donor[1])
+
+def print_donor():
+    donors = [x[0] for x in donor_db]
+    donors_list = "\n".join(donors)
+    print(donors_list)
+
+def print_email(name, amount):
+    letter = '''
+        "Dear {name},
```
hosungs: Minor: Did you want the leading double quote char? Also, there'll be leading spaces on every line, and not sure if that's what you wanted? In fact, that's kind of a common situation, so I remember there's a Python package that strips the leading spaces in the `'''...'''` strings, but I forgot the name... Maybe `dedent`? You may try to look it up... Not a major issue, though.

hosungs (commented):
Looks good overall! I'll leave it up to TA for any further comments & merging. Feel free to push any more commits if you have more. Thanks!

students/yushus/session03/mailroom.py
yushusongmsazure: Thank you so much Hosung! Yeah, the leading double quote is a mistake and will remove that. My initial thought is to make the email content itself "stand out" from the rest, hence chose to use ' ' '

students/yushus/session03/list_lab.py
yushusongmsazure: Yeah, I should not modify the original list while iterate through it and it indeed caused some issue in later section. I forgot to change this one :-)

students/yushus/session03/list_lab.py
yushusongmsazure: I see. Yeah, changed

students/yushus/session03/list_lab.py
yushusongmsazure: Changed

hosungs (approved):
Great that you followed up on my comments! Approving/merging this PR for the next assignment. TA grading comments may be left on Canvas. Thanks!

PR #39, opened 2018-10-24
Session03 assignments

motians:


students/shawn_degraw/session03/list_lab.py
```diff
@@ -0,0 +1,85 @@
...
+print("**Series 3**")
+fruit_list3 = fruit_list[:]
+new_list = []
+for fruit in fruit_list3:
+    while True:
+        answer = input("Do you like {}? ".format(fruit.lower()))
+        if answer == "no":
+            break
+        elif answer == "yes":
+            new_list.append(fruit)
```
hosungs, line 66: Ah, you added liked fruits to a new/empty list. I thought the intention was to delete disliked fruits from a copy of the original list, practicing iteration over a copy of the original list (because you don't want your loop to be affected by removal on the original list). Just FYI, and no need to modify this code.

students/shawn_degraw/session03/mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+#donor_db definition and default values
+donor_db = [("John Smith", [500.00, 150.00, 20.00]),
+            ("Jane Doe", [340.00, 30.00, 200.00]),
+            ("Jason Bourne", [240.00, 140.00]),
+            ("John Wick", [1000.00]),
+            ("GI Jane", [150.00, 60.00])
+           ]
+
+#Thank you letter template
+thankyouletter = "\n".join(("","Dear {name},","","Thank you for your generous donation of ${amount:.2f} to our cause.","Your donations help keep Python great!","","Sincerely","","The Python Project",""))
```
hosungs, line 13: Minor: Consider breaking to new lines after each comma, or you may want to use a `'''...'''` string for a multi-line string like this. Just to increase the code readability.

students/shawn_degraw/session03/mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+    """Prints a list of all donors in the db"""
+    for name in donor_db:
+        print(name[0])
+
+def handlename(namechoice):
+    """Determines if name provided is existing donor or new donor. Calls function to add donation to DB and calls function to print thank you letter
+    :param namechoice: name of the donor entered by user
+    """
+    #getting the name index starting from 1 to use truthiness in if statement
+    nameidx = [idx for idx, nametuple in enumerate(donor_db,1) if nametuple[0] == namechoice]
```
hosungs, line 48: Great list comprehension!

students/shawn_degraw/session03/slicing_lab.py
```diff
@@ -0,0 +1,47 @@
...
+    """ Provide any type of sequence and return the first and last items exchanged """
+    return seq[-1:] + seq[1:len(seq)-1] + seq[:1]
+
+def remove_every_other(seq):
+    """ Return every other item in the provided sequence """
+    return seq[::2]
+
+def double_minus_four(seq):
+    """ remove first 4 and last 4 and then return every other item """
+    return seq[4:len(seq)-4:2]
```
hosungs, line 11: Minor: No need for `len(seq)-4`. Just use `-4`.

students/shawn_degraw/session03/strformat_lab.py
```diff
@@ -0,0 +1,56 @@
...
+#Task Two
+print("\nTask Two")
+value1=2
+value2=123.4567
+value3=10000
+value4=12345.67
+#?? trying to get significance of 3
+getcontext().prec = 3 
+#??not sure about 3 significant figures for value4
+print(f"File_{str(value1).zfill(3)} :  {round(value2,2)}, {Decimal(value3):.2e}, {Decimal(value4):.2e}")
```
hosungs, line 16: Minor: Looks like you've done a lot of research on this! I never knew `.zfill()` method on a string. I think there must be some related f-string syntax that supports 0-padding, but I'm too lazy to look it up... Other things as well, like `round(...)`. Just FYI...

students/shawn_degraw/session03/strformat_lab.py
```diff
@@ -0,0 +1,56 @@
...
+
+#Task Three
+#??No commas? 
+print("\nTask Three")
+def formatter(in_tuple):
+    form_string = f'the {str(len(in_tuple))} numbers are:  {("{:2}" * len(in_tuple))}'
```
hosungs, line 22: Minor: Nice construction of a variable-length formatting string at the end! Regarding the first part, I wonder if you had to use `str(...)`? Did it not work without it? I think it should have worked, and you may want to avoid unnecessary `str` construction (which is automatically done by the Python runtime for any f-string anyway).

students/shawn_degraw/session03/strformat_lab.py
```diff
@@ -0,0 +1,56 @@
...
+
+#Task Four
+print("\nTask Four")
+print("{3:02} {4:02} {2:02} {0:02} {1:02}".format(4, 30, 2017, 2, 27))
+
+#Task Five
+print("\nTask Five")
+tasklist = ['oranges', 1.3, 'lemons', 1.1]
+
+print(f"The weight of an {tasklist[0][0:-1]} is {tasklist[1]} and the weight of a {tasklist[2][0:-1]} is {tasklist[3]}")
```
hosungs, line 36: Minor: No need to put `0` for the starting index of a slice. That is, `tasklist[0][:-1]` is just fine. 0 is the default starting index when it's not specified.

hosungs (approved):
Thanks for your hard work! Approving/merging this PR for your next assignment. TA may leave more comments on Canvas for his actual grading.

PR #40, opened 2018-10-24
Session 3 exercises complete for Dave Arasim

davidarasim:


students/dave_arasim/session03/list_lab.py
```diff
@@ -0,0 +1,188 @@
...
+print()
+
+fruit_list_ct = len(fruit_list)
+
+print('Please enter the fruit number to display (1 - ', fruit_list_ct, '): ', sep='', end='')
+fruit_num = input()
+try:
+    fruit_num = int(fruit_num)
+except ValueError:
+    fruit_num = ''
```
hosungs, line 30: Great to see you are already using exception handling! But it looks like this better be combined with the following `if`-`else` block. No need to separate these two out like this. That is,

```
try:
    fruit_num = int(fruit_num)
    if (fruit_num >= 1) and (fruit_num <= fruit_list_ct):
        print(fruit_num, ': ', fruit_list[(fruit_num-1)], sep='')
    else:
        print('That fruit number: ', fruit_num, ' is out of range', sep='')
except ValueError:
    print('Sorry, but that is not a number')
```
Well, actually this approach contains more code in the `try` block then necessary, so it might not be considered good yet. Maybe we should use the Python-specific `else` construct like the following?:
```
try:
    fruit_num = int(fruit_num)
except ValueError:
    print('Sorry, but that is not a number')
else:
    if (fruit_num >= 1) and (fruit_num <= fruit_list_ct):
        print(fruit_num, ': ', fruit_list[(fruit_num-1)], sep='')
    else:
        print('That fruit number: ', fruit_num, ' is out of range', sep='')
```
Frankly, I haven't used `else` on `try`-`except` that much, so I'll need to research more. We'll learn all these next week. Great to see you are looking ahead, David!

students/dave_arasim/session03/list_lab.py
```diff
@@ -0,0 +1,188 @@
...
+
+print('Fruit to delete: ', end='')
+fruit_name = input()
+print()
+
+if fruit_name == '':
+    print('Nothing input...delete cancelled.')
+else:
+    fruit_found = False
+    for i,this_fruit in reversed(list(enumerate(fruit_list))):
```
hosungs, line 95: Hmm... I'd say it's an interesting way of solving the `.pop()` indexing issue, but I suppose you could use `fruit_list.remove(fruit_name)`? Of course, it should first satisfy the problem requirements (which are vague to me at this time... sorry), but I don't think the intention of this problem is to apply a nontrivial workaround like this... Just a thought and comment. No need to fix this code.

students/dave_arasim/session03/mailroom.py
```diff
@@ -0,0 +1,155 @@
...
+            find_donor(thank_you_choice)
+            quit_thank_you = True
+
+def show_donor_list():
+    donor_list   = 'List of Donors'
+    donor_list_w = 26
+
+    #Donor list header
+    list_nom = (donor_list, donor_list_w)
+    print('{:{}}'.format(*list_nom))
```
hosungs, line 78: Hmm... quite interesting... I never knew a format string can have nested `{}` units! Interesting to see this working, but I think you can just specify the width in the format string (`'{:26}'.format(donor_list)`). I suppose you wanted to be able to avoid hard-coding `26` in multiple places, and if that was the intention and if this is the way to implement that, kudos! And thanks for teaching me something new and interesting!

students/dave_arasim/session03/mailroom.py
```diff
@@ -0,0 +1,155 @@
...
+
+    print()
+
+def find_donor(thank_you_choice):
+    for this_donor in range(len(donor_db)):
+        this_name = (donor_db[this_donor])[0]
+
+        if thank_you_choice == this_name:
+            break
+    else:
```
hosungs, line 97: Great to see you are utilizing the Python's unique `else` block in `for`!

hosungs (approved):
Great work! Approving/merging this PR for your next assignments. TA may provide more feedback on Canvas with his score grading. Thanks!

PR #41, opened 2018-10-24
PR for Session03 Files

bmnguyen:
So far I am only able to get two files done.  The rest will come as I have it done, perhaps a bit latter than I wish I could have them completed.


students/brandon_nguyen/session03/list_lab.py
```diff
@@ -0,0 +1,77 @@
...
+#SERIES 2 part
+    Series2_lst=Series1_lst[:]
+    print(Series2_lst)
+    Series2_lst.pop()
+    print(Series2_lst)
+    fruitInputDel = input("Please enter another fruit to delete and press enter >: ")
+  
+    for name in Series2_lst:
+        if name == fruitInputDel:
+            Series2_lst.remove(fruitInputDel)
```
hosungs, line 51: You are modifying the list that's being iterated (`Series2_lst`), and the loop's actual behavior might not be what you intended. For example, if `Series2_lst` contains duplicate items that are to be deleted (`name == fruitInputDel`) in consecutive indices, you may find that not all of those duplicate items are deleted with this code.

To avoid that, always iterate over a copy of the list that may be modified. To do that, you can replace the `for` loop header (line 49) to `for name in Series2_lst[:]:`.

students/brandon_nguyen/session03/mailroom.py
```diff
@@ -0,0 +1,114 @@
+import sys
+import unittest
+from decimal import Decimal
+
+#!/usr/bin/env python3
```
hosungs, line 5: This line should be always the first line of any `.py` file (if you want to put it), even before all `import` statements.

students/brandon_nguyen/session03/mailroom.py
```diff
@@ -0,0 +1,114 @@
...
+          ">>> "))
+
+def send_ty():
+    while True:
+        response = input("\nPlease enter donor full names or \"list\". Q to return to main menu: " )
+        if response.strip() == 'list':
+            list_donors()
+            break
+        elif response.strip() == 'Q':
+            main()
```
hosungs, line 39: Not a good use of recursion, because your `main()` already repeats with `while True:`. Just simply `break` out of the `while True:` loop here, or you can also simply `return` from here immediately.

students/brandon_nguyen/session03/mailroom.py
```diff
@@ -0,0 +1,114 @@
...
+#The script should prompt the user (you) to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.
+
+#think about a main menu function that can be called any where.
+def main_menu():
+    promptTxt = "\n".join(("Welcome to the mailroom program!",
+          "Please choose from below 3 options:",
+          "1 - Send a Thank You",
+          "2 - Create a Report",
+          "3 - Quit",
+          ">>> "))
```
hosungs, line 30: Not sure why you need this function. You are only redefining the `promptTxt` variable with the same content.

In fact, I don't see this function called from anywhere (from a cursory look, so maybe I'm wrong), so this function better be just removed.

students/brandon_nguyen/session03/mailroom.py
```diff
@@ -0,0 +1,114 @@
...
+            main()
+        else:
+            for name in donor_db:
+            #print(name[0])
+                if name[0] == response.strip():
+                    indx = donor_db.index(name)
+                    while True:
+                        donation = input("Please enter donation amount for " + name[0] +". Q to quit. :>>")
+                        if donation.strip() == 'Q':
+                            main()
```
hosungs, line 48: Again, not a good use of recursion. Since you are already in another nested loop, `break` may not work. However, you are in a function, and you just need to get out of the function in this case, so a simple `return` here will just work.

students/brandon_nguyen/session03/mailroom.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    #Currently need more work on the format as 0.0 problem not yet solved.
+    print()
+    newList = sort_sum() #creating a new list with sorted total in reverse already for easy printing.
+    header_string = "Donor Name              |  Total Given | Num Gifts | Average Gift"
+    line = '-'*len(header_string)
+    print(header_string)
+    print(line)
+    for donor, total, num, avg in newList:
+        print( donor, "\t\t$ ",str(total[0]).rjust(13), str(num[0]).rjust(5), "\t   $ ", str(avg[0]).rjust(10) )
```
hosungs, line 69: It's recommended not to use `\t`, and this formatting better be achieved using various `"...{...}...{...}...".format(...)` approach (or Python 3.6+'s new f-strings as well).

students/brandon_nguyen/session03/mailroom.py
```diff
@@ -0,0 +1,114 @@
...
+      
+def sort_sum():
+    #creating a new list with computed value for easy printing.
+    """
+    This function returned a sorted list of by on order amount3
+    """
+    dbSum = []
+    for name, donation in donor_db:
+        dbSum.append((name,[round(sum(donation),2)],[len(donation)],[round(sum(donation)/len(donation),2)]),)
+    return sorted(dbSum, key=lambda donor: donor[1], reverse=True)
```
hosungs, line 80: Great to see you are using `lambda` already!

students/brandon_nguyen/session03/slicing_lab.py
```diff
@@ -0,0 +1,73 @@
...
+import unittest
+
+#slicing lab in lession 3
+#Student Brandon Nguyen
+
+"""
+Exchange first and last sequence.
+"""
+def exchange_first_last(seq):
+    return seq[-1:]+seq[1:len(seq)-1]+seq[0:1]
```
hosungs, line 11: Minor: No need to use `len(seq)-1`. Just use `-1`, and Python runtime will automatically convert it to `len(seq)-1`.

Likewise, no need to specify `0` as the starting index. Python runtime will set the starting index to 0 if none is specified.

hosungs (approved):
Approving/merging this PR for your next assignment. Great to see your hard work!

students/brandon_nguyen/session03/list_lab.py
bmnguyen: Thank you for the great way of creating a copy on the fly.

students/brandon_nguyen/session03/mailroom.py
bmnguyen: got it!

students/brandon_nguyen/session03/mailroom.py
bmnguyen: I was thinking of something but never removed it.

students/brandon_nguyen/session03/slicing_lab.py
bmnguyen: Thank you!

PR #42, opened 2018-10-24
Hello 

aminavi:


students/aminavi/session03/list_lab.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
+
+list = ["Apple","Pears","Oranges","Peaches"]
```
hosungs, line 3: `list` is a built-in type name (for Python lists!), so it's not recommended to use it as your own variable name. Consider using something else, like `lst` or just `l`.

students/aminavi/session03/list_lab.py
```diff
@@ -0,0 +1,54 @@
...
+for i in list3:
+    like_fruit = input('do you like ' + i.lower() + '? ')
+    while like_fruit.lower() not in ('no', 'yes'):
+        like_fruit = input('do you like ' + i.lower() + '? ')
+    if like_fruit.lower() == 'yes':
+            result3.append(i.lower())
+print(result3)
+
+
+result4 =[w[::-1] for w in list]
```
hosungs, line 51: Great to see you are already using list comprehension!

students/aminavi/session03/mailroom.py
```diff
@@ -0,0 +1,63 @@
...
+thank_you_action = "Send a Thank You"
+create_a_report = "Create a Report"
+quit_action = "quit"
+
+def send_thank_you():
+    name = ''
+    while not name or name.lower() == 'list':
+        name = input("Please enter the full name: ")
+        if name.lower() == 'list':
+            print(' '.join(x[0] for x in donor_list))
```
hosungs, line 15: Cool way to list all donor names in one line!

students/aminavi/session03/mailroom.py
```diff
@@ -0,0 +1,63 @@
...
+
+    donation = int(input("Please enter the donation amount: "))
+    for i in range(len(donor_list)):
+        if name.lower() ==  donor_list[i][0].lower():
+            donor_list[i][1].append(donation)
+            return
```
hosungs, line 21: Hmm... If we return here, what about a thank you letter when the donor DB already has an entry for the entered donor name `name`?

I suppose you were trying to solve the case when the entered donor name `name` is not in the donor DB (line 22 case), but it probably should have been fixed by some logic like the following?:
```
    for i in range(len(donor_list)):
        if name.lower() ==  donor_list[i][0].lower():
            donor_list[i][1].append(donation)
            break
    else:
        donor_list.append((name, [donation]))
```
, using Python's unique `else` block in `for`...

students/aminavi/session03/mailroom.py
```diff
@@ -0,0 +1,63 @@
...
+        response = input(prompt)  # continuously collect user selection
+        if response == "1":
+            send_thank_you()
+        elif response == "2":
+            create_report()
+        elif response == "3":
+            print('Bye')
+            sys.exit()
+        else:
+            print("Not a valid option!")
```
hosungs, line 63: It's good to have a separate `main()` function for lines 52-63, and just call `main()` from the `if __name__ == '__main__':` block, rather than putting all the code here like this. That's just a good convention to follow. Just FYI.

students/aminavi/session03/slicing_lab.py
```diff
@@ -0,0 +1,32 @@
+seq = [4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21]
+str = "lets do something with this"
```
hosungs, line 2: `str` is another Python built-in type name (for any Python strings!), so it's not recommended to use it as your own variable name. Basically, you won't be able to do something like `str(num)` for a number object `num` once you redefine the name `str` like this. Make sure to use different names in the future.

students/aminavi/session03/slicing_lab.py
```diff
@@ -0,0 +1,32 @@
...
+# seq
+def exchange_first_last(x):
+    return x[-1:]+ x[1:-1]+ x[:1]
+# print(exchange_first_last(str))
+assert exchange_first_last(seq) == [21,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,4]
+assert exchange_first_last(str) == 'sets do something with thil'
+
+def return_everyother(x):
+	return x[0::2]
```
hosungs, line 11: Minor: No need to give 0 as the starting index, because it's the default when none is specified (and Python favors automatic defaults over explicitly specifying them).

students/aminavi/session03/slicing_lab.py
```diff
@@ -0,0 +1,32 @@
...
+
+def elemets_reversed(x):
+    return x[::-1]
+# print(elemets_reversed(str))
+assert elemets_reversed(seq) == [21, 20, 19, 18, 17, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4]
+assert elemets_reversed(str) == 'siht htiw gnihtemos od stel'
+
+def rearrange(x):
+    thrd = len(x)//3
+    return x[-thrd:] + x[0:thrd + thrd]
```
hosungs, line 29: Need to be careful with the integer division. I suspect you might lose some items if `len(x)` is a number that's not evenly divisible by 3 (like 5, 8, ...). Let's say `len(x)==5`. Then `thrd == 1`, and you will return `x[-1:] + x[0:2]`, which will be just `[x[-1], x[0], x[1]]`, which is `[x[4], x[0], x[1]]`, which misses `x[2]` and `x[3]`. This will need to be fixed correctly.

hosungs (approved):
Approving/merging this PR for your next assignment. TA may leave more comments on Canvas with his score grading. Great to see your hard work!

PR #43, opened 2018-10-24
session03 homework submission

jjakubiak:


students/jjakub/session03/list_lab.py
```diff
@@ -0,0 +1,49 @@
...
+mod_list = list_fruit[:]
+for frt in list_fruit:
+    user_pref = input("Do you like " + frt.lower() + "? ")
+    while user_pref.lower() not in ("yes", "no"):
+        user_pref = input("Please answer yes or no: ")
+    if user_pref.lower() == 'no':
+            mod_list.remove(frt)
+print(mod_list)
+
+rev_lst = [frt[::-1] for frt in list_fruit]
```
hosungs, line 46: Great to see that you are already using list comprehension!

students/jjakub/session03/mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+
+def sort_key(donor_db):
+    return sum(donor_db[1])
+
+def send_thanks():
+    name = ''
+    while not name or name.lower() == 'list':
+        name = input("Please enter the full name: ")
+        if name.lower() == 'list':
+            print(' '.join(x[0] for x in donor_db))
```
hosungs, line 25: Another list comprehension used to print all donor names in one line! Cool!

students/jjakub/session03/mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+
+    donation = int(input("Please enter the donation amount: "))
+    for i in range(len(donor_db)):
+        if name.lower() ==  donor_db[i][0].lower():
+            donor_db[i][1].append(donation)
+            return
```
hosungs, line 31: Hmm, if you return from here, then what about the thank you letter for the found donor?

I suppose you may want a logic like the following, using Python's unique `else` block for `for`?:
```
    for i in range(len(donor_db)):
        if name.lower() ==  donor_db[i][0].lower():
            donor_db[i][1].append(donation)
            return
    else:
        donor_db.append((name, [donation]))
```
Of course we can avoid all this complication next week with `dict`-based donor DB...

students/jjakub/session03/slicing_lab.py
```diff
@@ -0,0 +1,38 @@
...
+
+def mid_everyother(seq):
+    return seq[4:-4:2]
+
+def reverse(seq):
+    return seq[::-1]
+
+def thirds(seq):
+    thrd = len(seq)//3
+    return seq[-thrd:] + seq[0:thrd + thrd]
```
hosungs, line 19: I need to refer you to my same comment on another PR for this code: [here](https://github.com/UWPCE-PythonCert-ClassRepos/Au2018-Py210B/pull/42#discussion_r228295093). It looks like you two are somehow sharing incorrect code with each other, and that's not very good. Please do work on your code by yourself. This will have to be strictly enforced from the next assignment on.

/cc @Saravji and @aminavi FYI as well.

hosungs (approved):
Approving/merging this PR for your future assignment. Please make sure to check my comments and ask if you have any questions.

students/jjakub/session03/slicing_lab.py
jjakubiak: Thanks for the feedback. We were collaborating on the homework during the in class work sessions. I didn't realize it would be an issue. Apologies.

hosungs: Collaboration is fine and not an issue, but sharing code is discouraged, and especially sharing incorrect code may raise concerns. When collaborating, please make sure to understand what collaborated code means and how it works.

PR #44, opened 2018-10-24
Session03 Assignment

ejackie:


students/ejackie/session03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+
+print("\n----- Series 3 -----\n")
+
+fruits_iteration = fruits[:]
+print("The list has the following", len(fruits), "items:", fruits)
+for i, item in enumerate(fruits_iteration):
+    fruits[item] = item.lower()
+    answer = input("Do you like " + fruits[item] + "? ")
+    if(answer == "no"):
+        del fruits[item]
```
hosungs, line 52: Hmm... I may be mistaken, but does this work? `item` would be a string here, and `fruits` is a list, not a dict, so passing a string to a list would cause an exception (`TypeError`), no?
```
In [1]: l = ['a', 'b', 'c']

In [2]: del l['a']
------------------------------------------------------------------------
TypeError                              Traceback (most recent call last)
<ipython-input-2-6a3b5b15d27c> in <module>()
----> 1 del l['a']

TypeError: list indices must be integers or slices, not str

In [3]:
```

students/ejackie/session03/mailroom.py
```diff
@@ -0,0 +1,88 @@
...
+          "Please choose from below options:",
+          "1 - Send a Thank You",
+          "2 - Create a Report",
+          "3 - Exit",
+          ">>> "))
+
+# Looking up the donor and add them to database if not exist
+def send_a_thank_you(donor_db):
+    name = input("Please enter the full name of the donor: ")
+    if(name == "list"):
```
hosungs, line 29: Minor: No need to put `(`, `)` around Python's `if`/`while` conditions (and it's not recommended).

students/ejackie/session03/mailroom.py
```diff
@@ -0,0 +1,88 @@
...
+    print('{:5}{:70}'.format('Dear', name, ''))
+    print()
+    print('{:60}{:20}'.format('Thank you very much for your generous donation of: ', amount, ''))
+    print()
+    print('{:60}{:20}'.format('Your help is very appreciated. Have a wonderful day!', ''))
+    print()
+    print('{:60}{:20}'.format('Your Sincerely,', ''))
+    print('{:60}{:20}'.format('The Mailroom', ''))
+    print('\n---------- End of the email ----------\n')
+    exit
```
hosungs, line 57: ? I never knew Python has keyword `exit`?

Wow, I'm learning something new everyday from student's submissions... :) Found [this](https://stackoverflow.com/questions/37215406/exit-is-not-a-keyword-in-python-but-no-error-occurs-while-using-it) and not sure what the effect of `exit` here would be.

No need to put anything like this here, if you just need the function to return to its caller (because here's the end of the function anyway). Do not even put `return` (which is a valid/right Python keyword to return from a function) here in that sense.

students/ejackie/session03/slicing_lab.py
```diff
@@ -0,0 +1,69 @@
...
+
+    a_new_sequence = seq[::-1]
+
+    return a_new_sequence
+
+def new_order_last_third_first_third_middle_third(seq):
+    """ with the last third, then first third, then the middle third in the new order """
+
+    # number of items for a third of the sequence
+    num_third = int(len(seq)/3)
```
hosungs, line 39: Minor: `len(seq)//3` may be preferred (over `int(len(seq)/3)`). Note that it's a Python 3 change. In Python 2, `3/2 == 1`, but in Python 3, `3/2 == 1.5`, and `3//2 == 1`.
```
hosungs@DESKTOP-GJT06Q0:/mnt/c/gitroot/Au2018-Py210B (master)$ python
Python 2.7.15rc1 (default, Apr 15 2018, 21:51:34)
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 3/2
1
>>>
hosungs@DESKTOP-GJT06Q0:/mnt/c/gitroot/Au2018-Py210B (master)$ python3
Python 3.6.6 (default, Sep 12 2018, 18:26:19)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 3/2
1.5
>>> 3//2
1
>>>
hosungs@DESKTOP-GJT06Q0:/mnt/c/gitroot/Au2018-Py210B (master)$
```

hosungs (approved):
Approving/merging this PR for your next assignments. TA may leave more comments on Canvas in his score grading. Great to see your hard work!

students/ejackie/session03/list_lab.py
ejackie: It actually doesn't work and got the same error. Logic has been updated to avoid this issue.

students/ejackie/session03/mailroom.py
ejackie: Thanks for pointing that out! Removed ( ) from the if statements.

students/ejackie/session03/mailroom.py
ejackie: Got it. Removed exit. Thanks!

students/ejackie/session03/slicing_lab.py
ejackie: Thanks for the suggestion! Updated the code with len(seq)//3.

PR #46, opened 2018-10-26
Revised version of Lesson3_activity

nallarun10:
Thanks for correcting me. Made corrections.

Arun

### DIFF
--- a/students/arun_nalla/session03/Lesson3_Mailroom.py
+++ b/students/arun_nalla/session03/Lesson3_Mailroom.py
@@ -1,62 +1,60 @@
 #!usr/bin/en python
-donors_list = [('Aron 123', [10,5]), ('Becky', [3,30]), ('Charlie',[10,26]),('Thomas',[22,9,4,5]),('Nancy',[1,2,3])]
-donors_name = [x for x,y in donors_list]
-donors_number = {x: len(y) for x,y in donors_list}
-donors_total = {x:sum(y) for x,y in donors_list}
+donors_list = [('Aron King', [1,2]), ('Becky Patty',[3000,1000]), ('Charlie Lee',[666,222]),('Thomas Dave',[202,90,450]),('Nancy Crow',[100,200,300])]
 
-def thank_you(seq):
-    name_input = input("Enter a name from the list: ").title()
-    if name_input in donors_name:
-        print("\nDear Mr.{},\nThank you for donation of ${:.2f} for XYZ Charity Organisation. Its a kind initiative from"
-        " you for a nobel cause.\n\nWith Regards\n\nXYZ\n".format(name_input, donors_total[name_input]))
-    else:
-        print ('Name entered is not in the donor list')
-        thank_you(seq)
-def donors_thanks(seq):
-    name_input = input("Enter the donor of your interest: ").title()
+def thanks_email(seq):
+    name_input = input("Enter full name of a donor\n or 'Q' to quit>>>").title()
     if name_input == 'List':
-        print (donors_name)
-        thank_you(seq)
-    elif name_input in donors_name:
-        print("Dear Mr.{},\nThank you for donation of ${:.2f} for XYZ Charity Organisation. Its a kind intiative from"
-        " you for a nobel cause.\n\nWith Regards\n\nXYZ\n".format(name_input, donors_total[name_input]))
-    elif name_input !=seq:
-        seq.append(name_input)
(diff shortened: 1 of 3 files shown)


students/arun_nalla/session03/Lesson3_Mailroom.py
```diff
@@ -1,62 +1,60 @@
...
-    elif name_input in donors_name:
-        print("Dear Mr.{},\nThank you for donation of ${:.2f} for XYZ Charity Organisation. Its a kind intiative from"
-        " you for a nobel cause.\n\nWith Regards\n\nXYZ\n".format(name_input, donors_total[name_input]))
-    elif name_input !=seq:
-        seq.append(name_input)
-        print("Dear {},\nI am writing regarding the charity program 'XYZ' for 'ABC' cause. \nIf you are interested"
-              " please reply back. Your help will be appreciated.\n\nThank you\nXYZ\n".format(name_input))
+        name_list = [x for x,y in seq]
+        print (name_list)
+        thanks_email(seq)
```
hosungs: Still, not a good use of recursion. Imagine the use enters `List` a few times. Then this `thanks_email()` calls will pile up on the call stack, and when the top `thanks_email()` call returns, your for loop starting from line 12 will get executed again (because the earlier `thanks_email()` function execution has to continue, leading to line 12), and you'll get unexpected results.

Consider rewriting this using a `while` loop.

students/arun_nalla/session03/Lesson3_Mailroom.py
```diff
@@ -1,62 +1,60 @@
...
+    elif name_input =='Q':
+        main(seq)
```
hosungs: Ditto. Not a good use of recursion. In this case, you can simply `return`, instead of calling `main()`. 

Remember when this `main()` call returns (which does, because your `quit_program()` doesn't do `sys.exit()`, the execution of this function has to continue (to line 12, because this function is not done yet), and that'll give you unexpected results.

students/arun_nalla/session03/strformat_lab.py
```diff
@@ -11,15 +11,15 @@
 
 #task2
 # alternative formating
-s2 = '(file_00%s : %.2f, %.2e, %.2e)'%s
-print ('Task2')
+s2 = '(file_00%s: %.2f, %.2e, %.2e)'%s
```
hosungs: It looks like I missed this earlier. If `s[0]` is a number with more than 1 digit (e.g., `11`), your string will be something like `file_0011: ...`, which is not what was asked. You'll need to generate the left padding using a format string syntax, not hard-coding a fixed amount in the string like this.

Also recommended is to use `"...".format(...)`, rather than the old `"..." % (...)` style.

hosungs (commented):
Left a few more comments. Merging this PR for the next assignment.

@Saravji -- FYI again. Not sure how this will be reflected in your Canvas grading, but just letting you know. I think it's OK for you to continue grading on Canvas as planned, and maybe consider any later PRs like this as some make-up? I'll leave it up to you, or let me know if you'd like to discuss. Thanks!

students/arun_nalla/session03/Lesson3_Mailroom.py
```diff
@@ -1,62 +1,58 @@
...
-    elif name_input !=seq:
-        seq.append(name_input)
-        print("Dear {},\nI am writing regarding the charity program 'XYZ' for 'ABC' cause. \nIf you are interested"
-              " please reply back. Your help will be appreciated.\n\nThank you\nXYZ\n".format(name_input))
+#!usr/bin/env python
+donors_list = [('Aron King', [1,2]), ('Becky Patty',[3000,1000]), ('Charlie Lee',[666,222]),('Thomas Dave',[202,90,450]),('Nancy Crow',[100,200,300])]
 
+def thanks_email(seq):
+    for i, v in enumerate(seq):
+        while True:
```
hosungs, line 6: Hmm... I think these two loops are reversed. Actually, not just reversed, but the `for` loop shouldn't even apply if `name_input == 'List'`, don't you think?

The loop control here is totally flawed. You've got the `for` loop that iterates over all items in `seq`, but that works only when `name_input == 'List'`, which doesn't even apply when the user enters a correct/existing name. When the user enters a name that already exists in the database, your line 18 will get activated no matter what (assuming the entered name is not at index 0), which will make just append another entry in the list for the same name, and that's not what was asked.

Note that the `for` loop is to search the `seq` for the entered `name_input`, when it's not `'List'`. So you'll need to rewrite the loops correctly to get this fixed.

hosungs (commented):
Left a comment and merging this PR for the next assignment.

PR #47, opened 2018-10-26
turn in strformat

bmnguyen:
Please accept this new PR for the string format I am late to it.

### DIFF
--- a/students/brandon_nguyen/session03/strformat_lab.py
+++ b/students/brandon_nguyen/session03/strformat_lab.py
@@ -1,0 +1,50 @@
+import unittest
+from decimal import Decimal
+
+#!/usr/bin/env python3
+#Week3 Excercise string formating
+#Student: Brandon Nguyen - Au2018
+
+#task 1 and 2
+
+def strTask12():
+    strT=( 2, 123.4567, 10000, 12345.67)
+    #formt to produce 'file_002 :   123.46, 1.00e+04, 1.23e+04'
+    print("\'file_00{}:   {:.2f}, {:.2e}, {:.2e}\'".format(strT[0],strT[1],Decimal(strT[2]),Decimal(strT[3]))) #replace return/print to test
+
+#Task3 - Dynamically Building up format strings.  
+def formatter(in_tuple):
+    strFrm = "This is {:d}: {:d}"
+    #dynamically rebuild the strformat
+    for i in range(len(in_tuple)-1):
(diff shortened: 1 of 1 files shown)


students/brandon_nguyen/session03/strformat_lab.py
```diff
@@ -0,0 +1,50 @@
...
+#!/usr/bin/env python3
+#Week3 Excercise string formating
+#Student: Brandon Nguyen - Au2018
+
+#task 1 and 2
+
+def strTask12():
+    strT=( 2, 123.4567, 10000, 12345.67)
+    #formt to produce 'file_002 :   123.46, 1.00e+04, 1.23e+04'
+    print("\'file_00{}:   {:.2f}, {:.2e}, {:.2e}\'".format(strT[0],strT[1],Decimal(strT[2]),Decimal(strT[3]))) #replace return/print to test
```
hosungs, line 13: If `strT[0]` is a number with more than 1 digit (e.g., 11), your print out will be `file_0011: ...`, which is not what was asked.

You used `{:02d}` below in line 25, and you should use that here as well, to put 0 padding on the left with the fixed size width.

students/brandon_nguyen/session03/strformat_lab.py
```diff
@@ -0,0 +1,50 @@
...
+stp = ( 4, 30, 2017, 2, 27)
+#to print out '02 27 2017 04 30'
+print("\'{:02d} {:d} {:d} {:02d} {:}".format(stp[3],stp[4],stp[2],stp[0],stp[1]))
+
+print()
+#Task 5all about 5 f-string:
+def strTask5a():
+    lst = ['oranges', 1.3, 'lemons', 1.1]
+    #display The weight of an orange is 1.3 and the weight of a lemon is 1.1
+    text = f'The weight of an {(lst[0]).replace("s","")} is {lst[1]} and the weight of a {(lst[2]).replace("s","")} is {lst[3]}'
```
hosungs, line 32: `lst[0].replace("s","")` won't work as expected if the fruit name contains `s` (e.g., `strawberry`). Probably a slicing (to remove the last letter) might work better.

hosungs (commented):
Left a few more comments. Merging this PR for the next assignment.

@Saravji -- Not sure how you can include this new change in your grading (if no new files are submitted to Canvas), but just FYI...

PR #48, opened 2018-10-27
Revised Lesson3_Mailroom activity

nallarun10:


hosungs (commented):
Still some issue within your for loop...

students/arun_nalla/session03/Lesson3_Mailroom.py
```diff
@@ -2,26 +2,27 @@
...
+        else:
+            for i, v in enumerate(seq):
+                if name_input == seq[i][0]:
+                    donation_amount = float(input('Enter donation amount: '))
+                    seq[i][1].append(donation_amount)
+                    print("Dear {},\n\nThis message is regarding the recent donation of ${:.2f} you made to ZYX charity.\n"
+                    "We highly appreciate your kindness. Till date you have donate a total of ${:.2f}.\n\n"
+                    "With Regards\n\nXYZ\n".format(name_input, donation_amount,sum(seq[i][1])))
+                    return seq
+                elif name_input != seq[i][0]:
```
hosungs: I think there's still some issue in your loop control here.

If you enter a name that already exists, but not at the first index (that is, try enter something like `Becky Patty`), then your code won't append the new donation to the existing `Becky Patty`'s donation list, but it's going to create a whole new record for another `Becky Patty`, with a new donation list for that single donation entry, and append this tuple at the end of the `seq` list. I don't think that's what was asked.

Note that the `seq.append((name_input, [donation_amount]))` can be done only after all names were searched and no match was found. In other words, that should be done outside the `for` loop, not inside it like this. Because any `name_input` that doesn't match `seq[0][0]` (like `Becky Patty`, which matches `seq[1][0]`) will satisfy the `elif` condition here, execution will fall through here and behave as I stated above, not as expected.

Note that I'm just providing this comment out of just reading your code, not by executing it at all, so I might be wrong, but I'm pretty sure that's how your code would behave. When you test your code, please make sure that it actually works as intended, not just skimming through.

students/arun_nalla/session03/Lesson3_Mailroom.py
```diff
@@ -5,24 +5,26 @@ def thanks_email(seq):
...
                     seq[i][1].append(donation_amount)
                     print("Dear {},\n\nThis message is regarding the recent donation of ${:.2f} you made to ZYX charity.\n"
-                    "We highly appreciate your kindness. Till date you have donate a total of ${:.2f}.\n\n"
-                    "With Regards\n\nXYZ\n".format(name_input, donation_amount,sum(seq[i][1])))
+                            "We highly appreciate your kindness. Till date you have donate a total of ${:.2f}.\n\n"
+                            "With Regards\n\nXYZ\n".format(name_input, donation_amount, sum(seq[i][1])))
                     return seq
-                elif name_input != seq[i][0]:
+            for i, v in enumerate(seq):
+                if name_input != seq[i][0]:
```
hosungs, line 20: Getting closer, and I think this code should behave as expected. However, these lines 19-20 are not needed at all, because getting out of the previous `for` loop means that there was no matching name in the `seq`, so there's really no need to do another loop of comparisons at all.

hosungs (commented):
Left another comment, and merging/closing this PR for the next assignment.

PR #49, opened 2018-10-28
session04

jjakubiak:
Still working on session 04, not ready for review. Thanks.

### DIFF
--- a/students/jjakub/session03/slicing_lab.py
+++ b/students/jjakub/session03/slicing_lab.py
@@ -6,7 +6,8 @@ def first_last(seq):
     return seq[-1:]+ seq[1:-1]+ seq[:1]
 
 def everyother(seq):
-    return seq[0::2]
+    return seq[::2]
+    # Removed zero as the starting index because it is already the default
 
 def mid_everyother(seq):
     return seq[4:-4:2]
@@ -15,8 +16,9 @@ def reverse(seq):
     return seq[::-1]
 
 def thirds(seq):
-    thrd = len(seq)//3
-    return seq[-thrd:] + seq[0:thrd + thrd]
+    thrd = int(round(len(seq)/3,0)*2)
+    return seq[thrd:] + seq[:thrd]
+    # Replaced integer division with precision rounding to capture all items in strings that are not evenly divisible by 3
 
 assert first_last(a_vec) == [15,2,3,4,5,6,7,8,9,10,11,12,13,14,1]
 assert first_last(a_str) == "ghis is a strint"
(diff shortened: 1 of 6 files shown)


students/jjakub/session03/slicing_lab.py
```diff
@@ -15,8 +16,9 @@ def reverse(seq):
     return seq[::-1]
 
 def thirds(seq):
-    thrd = len(seq)//3
-    return seq[-thrd:] + seq[0:thrd + thrd]
+    thrd = int(round(len(seq)/3,0)*2)
+    return seq[thrd:] + seq[:thrd]
```
hosungs, line 20: This looks better (and probably correct), but the whole thing could have been simply as follows:
```
third = len(seq)//3
return seq[-third:] + seq[:third] + seq[third:-third]
```
Or maybe just `return seq[-third:] + seq[:-third]`. No need to change your code.

I'll leave this PR for your future commits for session 4. Thanks.

jjakubiak:
I committed the session04 trigram assignment. The program works when running in the interpreter but I can't get it to work from the cmd line. I believe the issue is the file path/name that is passed from the user prompt.

Also, I am still working on the second part of the mailroom assignment and will submit asap. 

Thanks!

students/jjakub/session04/trigram.py
```diff
@@ -0,0 +1,50 @@
...
+#     else:
+#         d[pair].append(follower)
+
+### create dictionary
+def build_trigrams(words, wordcnt):
+    d = {}
+    for i in range(len(words)-2):
+        pair = tuple(words[i:i + 2])
+        follower = words[i + 2]    
+        d.setdefault(pair,[]).append(follower)
```
hosungs, line 22: Minor: Good usage of `.setdefault()`! You may find `collections.defaultdict` more useful in the future... No need to change here.

students/jjakub/session04/trigram.py
```diff
@@ -0,0 +1,50 @@
...
+        trigrams.extend(d.setdefault(tuple(cur_pair),[]))
+
+    return trigrams
+
+if __name__ == "__main__":
+    trigrams = build_trigrams(words, wordcnt)
+    print(trigrams)
+
+
+filepath = '.\\students\\jjakub\\session04\\'
```
hosungs, line 41: I'm responding to your comment/remark that it doesn't work in the command line. Yes, that's likely the case, if you run your command line from your `session04` directory, not from the git repo root directory (in which case you'll need to run `python students/jjakub/session04/trigram.py`, which is quite inconvenient). If you'd like to avoid this inconvenience, you may start a new VS Code window, and then open your `students/jjakub/session04` folder, instead of the git repo.

hosungs (commented):
Sure, I added some comments related to your path issue, and hopefully it works for you.

And yes, keep working on the remainder and keep pushing your commits. Thanks.

hosungs:

PR #51, opened 2018-10-29
mailroom2, dict_lab, File_lab

guoguoying2013:
This PR contains:
Mailroom2
Dict_lab
File_lab

They are practice for session04
I will submit Trigrams later before due


students/YingGuo/session04/dict_set_lab.py
```diff
@@ -0,0 +1,89 @@
...
+
+"""Using the dictionary from item 1: Make a dictionary using the same keys but with the number of ‘t’s in each value as the value (consider upper and lower case?)"""
+
+dict2 = {}
+for x, y in dict1.items():
+    count = {"t":0}
+    for i in list(y):
+        if i == "t":
+            count["t"] = count["t"] +1
+    dict2[x] = count["t"]
```
hosungs: FYI: Lines 42-46 could be simply in one line: `dict2[x] = y.count('t')`

students/YingGuo/session04/dict_set_lab.py
```diff
@@ -0,0 +1,89 @@
...
+Display if s3 is a subset of s2 (False)
+and if s4 is a subset of s2 (True)"""
+
+s2 = set()
+s3 = set()
+s4 = set()
+
+for i in range(0,21):
+    if i%2 == 0:
+        s2.update([i])
```
hosungs, line 58: You used `.add()` in line 66, so why not using it here as well? Also in line 64?

students/YingGuo/session04/file_lab.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+
+def function1():
+    """a program which prints the full path for all files in the current directory, one per line"""
+    import pathlib
```
hosungs, line 5: Minor: All import statements are expected to be at the top of any script (after line 1).

students/YingGuo/session04/file_lab.py
```diff
@@ -0,0 +1,45 @@
...
+
+def function2():
+        """Write a program which copies a file from a source, to a destination (without using shutil, or the OS copy command).
+        Advanced: make it work for any size file: i.e. don’t read the entire contents of the file into memory at once.
+        This should work for any kind of file, so you need to open the files in binary mode: open(filename, 'rb') (or 'wb' for writing). Note that for binary files, you can’t use readline() – lines don’t have any meaning for binary files. Test it with both text and binary files (maybe jpeg or something of your choosing)."""
+        with open ("students.txt","r") as f_input, open("students_copy.txt","w") as f_output:
+                for line in f_input.readlines():
+                        f_output.write(line)
+ 
+# are we going to have example answer for lab?
```
hosungs, line 19: Yes.

students/YingGuo/session04/file_lab.py
```diff
@@ -0,0 +1,45 @@
...
+        print("The full path for {} is {}".format(f,f.absolute()))
+
+
+def function2():
+        """Write a program which copies a file from a source, to a destination (without using shutil, or the OS copy command).
+        Advanced: make it work for any size file: i.e. don’t read the entire contents of the file into memory at once.
+        This should work for any kind of file, so you need to open the files in binary mode: open(filename, 'rb') (or 'wb' for writing). Note that for binary files, you can’t use readline() – lines don’t have any meaning for binary files. Test it with both text and binary files (maybe jpeg or something of your choosing)."""
+        with open ("students.txt","r") as f_input, open("students_copy.txt","w") as f_output:
+                for line in f_input.readlines():
+                        f_output.write(line)
```
hosungs, line 17: FYI, this works only for text files. Try to achieve this for binary files as well...

students/YingGuo/session04/mailroom2.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+
+#session04, mailroom2
+
+from collections import defaultdict
+data = defaultdict(list)
+Sample = {"Tom":[100,200], "Tim":[3000], "Ken": [888,888,888], "Ted":[8000,5000,500], "Jon":[1000,6000]}
+for key in Sample.keys():
+    data[key] = Sample[key]
```
hosungs: Wondering why you had to create `Sample` separately and copy like this? I suppose it's partly because we didn't talk about how to initialize `defaultdict(list)` with values. You could have simply done the following:
```
data = defaultdict(list, {"Tom":[100,200], "Tim":[3000], "Ken": [888,888,888], "Ted":[8000,5000,500], "Jon":[1000,6000]})
```
for lines 6-9...

students/YingGuo/session04/mailroom2.py
```diff
@@ -0,0 +1,67 @@
...
+
+		elif user_input.lower() == "quit":
+			break
+		
+		else:
+			amount1 = int(input("how much would you like to donat?"))
+			data[user_input.capitalize()] += [amount1]
+			print("Thank you {} for your donation {}!".format(user_input.capitalize(), amount1))
+			return data
+			break
```
hosungs: Minor: No need to `break`, because it can never be reached (because previous line is `return`).

students/YingGuo/session04/mailroom2.py
```diff
@@ -0,0 +1,67 @@
...
+	print(string_title)
+	print("-"*len(string_title))
+
+	for x in report_list:
+		print("{:<12} ${:>12,.2f}  {:>12} ${:>12,.2f}".format(*x))
+
+def all_letter(data):
+    """Try to use a dict and the .format() method to produce the letter as one big template, the write into a txt file named with donor's name"""
+    for name in data.keys():
+        with open("{}".format(name),"w") as f:
```
hosungs: Good file writing code. Just wanted to mention that you might want to save these letter files in a subdirectory, not at the same directory... Besides, `"{}".format(name)` is redundant--it's just the same as `name`. You might have wanted to say something like `"letters/{}.txt".format(name)`? You'll need to create your subdirectory `letters` first, though...

hosungs (commented):
Left some comments for your information. Will leave this PR open for your remaining assignments. Thanks for your hard work!

students/YingGuo/session04/Trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+    trigrams = defaultdict(list)
+
+    for i in range(len(words)-2):
+        pair = tuple(words[i:i+2])
+        follower = words[i+2]
+        trigrams[pair] += [follower]
+    
+    print("This is the Trigrams:")
+    for i in trigrams.keys():
+        print(i, ": ", trigrams[i])
```
hosungs: Minor: Desirable `for` loop iteration for this kind of use case is as follows:
```
for k, v in trigrams.items():
    print(k, ": ", v)
```
At least you don't need to put `.keys()` when you iterate over keys only (so `for i in trigrams:` would have just worked the same). Just FYI, and it'll take some time to absorb all these details...

students/YingGuo/session04/Trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+    return trigrams
+        
+def build_text(trigram_dict):
+    key_lst = list(trigram_dict.keys())
+    print("This is the Tuple key list:\n", key_lst)
+    fst_item = random.choice(key_lst)
+    new_lst = [fst_item[0].capitalize(),fst_item[1]]
+    while len(new_lst) <=200:
+        key_tuple = (new_lst[-2],new_lst[-1])
+        if key_tuple in key_lst:
```
hosungs, line 32: Minor again: No need to create a separate list `key_lst` for this purpose. You could have just used `if key_tuple in trigram_dict:`, because `k in a_dict` is about whether the key `k` is in the `a_dict` or not.

hosungs (commented):
Added a few more comments. Good job with Trigrams!

hosungs (approved):
Looks like this PR should be now merged/closed for the next assignment, so merging/closing it. Thanks!

PR #52, opened 2018-10-31
PR for assignment 4

yushusongmsazure:
trigrams program still need to be polished


students/yushus/session04/mailroom2.py
```diff
@@ -0,0 +1,93 @@
...
+          "3 - Send letters to all donors",
+          "4 - quit",
+          ">>> "))
+
+def print_donor():
+    donors = donor_db.keys()
+    donors_list = "\n".join(donors)
+    print(donors_list)
+
+def format_letter():
```
hosungs, line 32: Minor: Looks like this function returns a fixed string (a formatting string), so not sure if it needs to be a function. The `letter` variable could have been declared at the global level (though, with some other/proper name), and just use the global variable in `print_email()`. Well, using global variables is kind of frowned upon, and in that case, you could receive `name` and `amount` as parameters to this function and return the actually constructed strings, not just a format string. Just FYI, and no need to change this.

students/yushus/session04/mailroom2.py
```diff
@@ -0,0 +1,93 @@
...
+            donor_db[name]=[]
+            break
+        else:
+            break
+
+    amount = float(input("How much do you want to donate? "))
+    donor_db[name].append(amount)
+    print_email(name, amount)
+
+    return
```
hosungs: Minor: No need to put `return` here (function will end anyway).

hosungs (commented):
Left a few comments. Looking good and keep going for polishing `trigrams.py` as you said!

hosungs (approved):
Glad to see all your hard work! Merging/closing this PR for the next assignment.

PR #55, opened 2018-10-31
 Session04 assignment. Mailroom2 done. Trigrams first version.

ejackie:


students/ejackie/session03/list_lab.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python
```
hosungs, line 1: Minor: If you run this script on a Linux platform, this change may start Python 2, instead of Python 3. Not sure if that's your intention? Also not sure how this line works in Windows with git-bash (if that's your environment)... Just a note, and if this still works fine, you can ignore this (but remember about the cross-platform compatibility).

students/ejackie/session04/mailroom2.py
```diff
@@ -0,0 +1,103 @@
...
+    while True: # loop forever until they quit.
+        response = input(prompt)
+        if dispatch_dict[response]() == "exit program":
+            break
+
+
+def send_a_thank_you():
+    name = input("Please enter the full name of the donor: ")
+    if name == "list":
+        for key in donor_dict.keys():
```
hosungs, line 38: Minor: No need to add `.keys()`. Just `for key in donor_dict:` will do (the same).

students/ejackie/session04/mailroom2.py
```diff
@@ -0,0 +1,103 @@
...
+            print(" ".join(key))
+        print()
+    elif tuple(name.split()) in donor_dict.keys():
```
hosungs, line 41: Same here. No need to put `.keys()` at the end.

students/ejackie/session04/trigrams_dict.py
```diff
@@ -0,0 +1,87 @@
...
+
+def read_in_data(filename):
+    """
+    read text from specified file
+
+    """ 
+    with open(filename, 'r') as f:
+        in_data = f.read()
+    f.closed
+    True
```
hosungs, line 24: Minor: I suppose these lines 22-23 were your debugging purpose only? No need for these lines otherwise. Just FYI.

students/ejackie/session04/trigrams_dict.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    trigrams = {}
+
+    # build up the dict here!
+
+    for i in range(len(words)-2):
+        pair = words[i:i + 2]
+        follower = words[i + 2]
+        if tuple(pair) in trigrams:
+            trigrams[tuple(pair)] = [trigrams[tuple(pair)], follower]
```
hosungs, line 55: Minor: Why not `trigrams[tuple(pair)].append(follower)`? Probably this has something to do with your line 52, where you assign `follower` as the value of `trigrams[tuple(pair)]`, not `[follower]`. You could also use `trigrams = defaultdict(list)` in line 47, after adding `from collections import defaultdict` at the top. Then there's even no need to check line 54 at all.

hosungs (commented):
Looking good. Keep going!

hosungs:

PR #56, opened 2018-10-31
Added Lesson4 activity

nallarun10:
Mailroom Part4 submitted
Still working on Trigram

Arun


nallarun10:
added Trigrams and dict_lab

students/arun_nalla/session03/Lesson3_Mailroom.py
```diff
@@ -16,14 +16,13 @@ def thanks_email(seq):
                             "We highly appreciate your kindness. Till date you have donate a total of ${:.2f}.\n\n"
                             "With Regards\n\nXYZ\n".format(name_input, donation_amount, sum(seq[i][1])))
                     return seq
-            for i, v in enumerate(seq):
-                if name_input != seq[i][0]:
-                    donation_amount = float(input('Enter donation amount: '))
-                    seq.append((name_input, [donation_amount]))
-                    print("Dear {},\nThis message is regarding the recent donation of ${:.2f} you made to ZYX charity.\n"
+            else:
```
hosungs, line 19: Good use of `else`, I think, but you don't even need this `else` at all, because the only case when the execution continues after the `for` loop is when there's no match. Therefore, `else` is already implied when the execution reaches here, so that's why I think `else` is not even needed here. Just FYI, no need to change.

students/arun_nalla/session04/Lesson4_Mailroom_Part2.py
```diff
@@ -0,0 +1,74 @@
+#!usr/bin/env python
+
+donors_list = [('Aron King', [1, 2]), ('Becky Patty', [3000, 1000]),
+               ('Charlie Lee', [666, 222]), ('Thomas Dave', [202, 90, 450]), ('Nancy Crow', [100, 200, 300])]
+
+donors_dict = {x: y for x, y in donors_list}
```
hosungs, line 6: Minor: FYI, `donors_dict` can be directly initialized as below, not through another list (`donors_list`):
```
donors_dict = { 'Aron King': [1,2], 'Becky Patty': [3000, 1000], ... }
```

students/arun_nalla/session04/Lesson4_Mailroom_Part2.py
```diff
@@ -0,0 +1,74 @@
...
+        else:
+            for name in donors_dict.keys():
+               if name_input == name:
+                   d = {}
+                   d['name'] = name
+                   donation_amount = int(input('Enter donation amount: '))
+                   donors_dict[name].append(donation_amount)
+                   d['donation'] = donation_amount
+                   print("Dear {name},\n\tThank you for generous donation of ${donation:.2f} to ZYX charity\nSincerely,\nZYX Team".format(**d))
+                   return
```
hosungs: Because you are now using a dict, there's no need to use a `for` loop to search for the entered name. You can simply say `if name_input in donors_dict:`, to check if the `donors_dict` contains an entry whose key matches `name_input` value. No `for` loop needed. In other words, you can change your lines 15 and 16 to `if name_input in donors_dict:`, and replace `name` in line 18, 20 & 22 with `name_input`.

students/arun_nalla/session04/Lesson4_Mailroom_Part2.py
```diff
@@ -0,0 +1,74 @@
...
+def quit_program(donors_dict):
+    print('Thank for using this script.\nHave a nice day\nBye\n')
+    exit()
+
+def email_donors(donors_dict):
+
+    for a in donors_dict:
+        message = ('Dear {},\n\n\tThank you for your very kind donation of ${:0.2f}.\n\tIt will be put '
+                   'to very good use.\n\n\t\tSincerely,\n\t\t-The Team'.format(a, sum(donors_dict[a])))
+        with open (r'C:\Users\Arun\Au2018-Py210B\students\arun_nalla\session04\Mailroom_messages\e_mail_{}.txt'.format(a), 'w') as f:
```
hosungs: No good to hard-code an absolute path like this... Just use 'Mailroom_messages/e_mail_{}.txt`.

If you are using VS Code and wanted the `Mailroom_messages` directory to be under your `session04` directory, just reopen your `session04` directory in a new VS Code window and that'll allow your code to work with the `Mailroom_messages` subdirectory under `session04`.

hosungs (commented):
Left more comments. Leaving this PR for further commits (trigram assignment).

nallarun10:
added Trigrams assignment

students/arun_nalla/session04/trigrams.py
```diff
@@ -0,0 +1,50 @@
...
+#words = 'I wish I may I wish I might'.split()
+def trigram_dict (words):
+    trilist = {}
+    for i in range(len(words)-2):
+        a = tuple(words[i:i+2])
+        b = words[i+2]
+        if a not in trilist:
+            trilist[a] = [b]
+        else:
+            trilist[a].append(b)
```
hosungs: Minor: Lines 8-11 can be reduced to `trilist.setdefault(a, []).append(b)`. Or even better is to use `defaultdict` and get rid of `.setdefault(...)`.

students/arun_nalla/session04/trigrams.py
```diff
@@ -0,0 +1,50 @@
...
+    word_list = ''
+    with open(filename, 'r') as file_name:
+        translator = str.maketrans('','', string.punctuation)
+        for lines in file_name.readlines():
+            words = lines.translate(translator)
+            word_list+=words
+    list_of_words = word_list.split()
+    return list_of_words
+import random
+import string
```
hosungs: Minor: Put `import` statements at the top of the source file.

students/arun_nalla/session04/trigrams.py
```diff
@@ -0,0 +1,50 @@
...
+def build_new_story(story):
+    keys_first = random.choice(list(trigram_dict(words)))
```
hosungs: Confusing usage of global `words`. Better receive it as a function input parameter.

students/arun_nalla/session04/trigrams.py
```diff
@@ -0,0 +1,50 @@
...
+    #print (keys_first, "key first")
+    my_list = list(keys_first)
+    #print (my_list, "new list")
+    my_list.append(random.choice(trigram_dict(words)[keys_first]))
```
hosungs: You are making `trigram_dict(words)` call many times (here, at line 25, 33, 34, anywhere else?). Note that every time you call `trigram_dict(words)`, Python runtime executes all statements in the function, which is basically reconstructing the same dict again and again. No need for such repeated/duplicate computation.

Call the function once, save the result to a variable, and use the variable whenever it's needed. In other words, put a statement `trigrams = trigram_dict(words)` after line 24, and replace every `trigram_dict(words)` occurrence in this function with `trigrams`.

hosungs (changes requested):
Left some comments. I'd like to see those addressed, so leaving this PR open.

hosungs (approved):

hosungs:

PR #57, opened 2018-10-31
Session 04 assignments

bmnguyen:
Currently only Mailroom 2 is finished.  However, the files are written to current dir and not a temp/.  Once again I am way behind with my schedule.

### DIFF
--- a/students/brandon_nguyen/session04/mailroom2.py
+++ b/students/brandon_nguyen/session04/mailroom2.py
@@ -8,11 +8,12 @@ from decimal import Decimal
 
 
 #global define data structure
-donor_db = [("Brandon Nguyen", [4500.33,350.87,300.05]),
-          ("Shawn DeGraw", [5500.00, 345.05,571.33]),
-          ("Jacqueline Lee",[4300.00,3200.00,230.13]),
-          ("Aidan Nguyen", [4300.00,200.00,238.23]),
-          ]
+#dictionary {peron:[donations]}
+donor_db = {"Brandon Nguyen": [4500.33,350.87,300.05],
+          "Shawn DeGraw": [5500.00, 345.05,571.33],
+          "Jacqueline Lee":[4300.00,3200.00,230.13],
+          "Aidan Nguyen" : [4300.00,200.00,238.23],
+        }
 promptTxt = "\n".join(("\nWelcome to the mailroom program!\n",
           "Please choose an options:\n\n",
           "1 - Send a Thank You to a single donor.",
@@ -42,10 +43,10 @@ def send_ty():
         elif response.strip() == 'Q':
             return
         else:
-            for name in donor_db:
+            for person in donor_db:
             #print(name[0])
-                if name[0] == response.strip():
-                    indx = donor_db.index(name)
+                if person == response.strip():
+                    #indx = donor_db.index(name)
                     while True:
                         donation = input("Please enter donation amount for " + name[0] +". Q to quit. :>>")
                         if donation.strip() == 'Q':
@@ -56,7 +57,8 @@ def send_ty():
                     print("\n\nHello {},\nThank you so much for the generious donation of {}!\n\n".format(name[0],float(donation)))
                     return
             else:
-                donor_db.append((response.strip(),[0.00])) #adding new name with $0 donation. avoiding div/0 issue
+                #donor_db.update((response.strip(),[0.00])) #adding new name with $0 donation. avoiding div/0 issue
+                addNew_person(response.strip(),donor_db)
 
 def send_ty_all():
     pass
@@ -98,11 +100,14 @@ def list_donors():
     for name in lst: print(name)
     print()
 #breaking down to sub function
-def update_donation(db):
-    """
-    
-    """
-    pass
+def addNew_person(newName, db):
+    newRcd = {newName:[0.00]}
+    db.update(newRcd)
+
+def update_donation(name,newDonation,db):
+    cur_donation = db.get(name)
+    db[name] = cur_donation.append(newDonation)    
+
 
 def quit_program():
     print("Thank you for trying mailroom!")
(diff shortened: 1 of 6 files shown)


students/brandon_nguyen/session04/trigrams_text.py
```diff
@@ -0,0 +1,60 @@
...
+    return trigrams
+
+
+if __name__ == "__main__":
+    trigramTest = {("I", "wish"): ["I", "I"],
+            ("wish", "I"): ["may", "might"],
+            ("may", "I"): ["wish"],
+            ("I", "may"): ["I"],
+            }
+    assert build_trigrams(words) == trigramTest
```
hosungs, line 72: Great to see your test-driven development practice!

students/brandon_nguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+#The script should prompt the user (you) to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.
+
+#think about a main menu function that can be called any where.
+def show_menu(promptxt, select_dict):
+    while True:
+        menu_selected = input(promptxt)
+        if select_dict[menu_selected]() == "Exit Menu":
+            break
```
hosungs, line 39: Great generalized function for arbitrary menu prompt/selections!

Now you learned exceptions, you may want to handle invalid choices here using `try: ... except KeyError: ...`! No need to do that change now. In fact, that's part of mailroom part 3 assignment.

students/brandon_nguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    This function returned a sorted list of by on order amount3
+    """
+    dblist = []
+    dbSum = []
+    #quick way to convert back to list to reuse existing code.
+    for i,j in donor_db.items():
+        dblist.append((i,j))
+    for name, donation in dblist:
+        dbSum.append((name,[round(sum(donation),2)],[len(donation)],[round(sum(donation)/len(donation),2)]),)
```
hosungs, line 70: Hmm... Not sure why you are making lists of single values as tuple members? Do you get errors when you remove all the `[` and `]` in this line 70?

students/brandon_nguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,143 @@
...
+    return sorted(dbSum, key=lambda donor: donor[1], reverse=True)
```
hosungs, line 71: Great to see your `lambda` usage, but since `donor[1]` would be a list of one number (sum), I'm not sure how it's compared. I guess list comparison follows item-by-item comparison, so that might be why this would work. Still, I think there's no need to create lists of single values in each `dbSum` item (tuple).

students/brandon_nguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,143 @@
...
+    
+def list_donors():
+    """
+    Basic Name of donor return in Order by First Name
+    """
+    lst = []
+    print()
+    for person in donor_db:
+        lst.append(person)
```
hosungs, line 80: Minor: Lines 77, 79-80 could be simplified to `lst = donor_db.keys()`.

Maybe you had an error in `lst.sort()`, when you used the `donor_db.keys()`. In that case, probably better to say `lst = sorted(donor_db.keys())`.

students/brandon_nguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,143 @@
...
+def update_donation():
+        input_person = input("\nPlease enter donor full names: " )
+        input_donation = float(input("Please enter donation amount for " + input_person + ":>> "))
+        #existing people add to donation
+        if input_person in donor_db:
+            new_Donation = donor_db.get(input_person)
+            new_Donation.append(input_donation)
+            donor_db[input_person] = new_Donation
+        else:
+            donor_db.update( {input_person :[input_donation]} )
```
hosungs, line 94: Minor: Lines 88-94 could be simplified to:
```
donor_db.set_default(input_person, []).append(input_donation)
```
Or better yet, use `defaultdict` for `donor_db` and even no need for `.set_default(...)`.

students/brandon_nguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,143 @@
...
+    e_frmt = {
+                  0:'Dear',
+                  1:'Thank you for your kind donation of ',
+                  2:'It will be put to very good use',
+                  3:'Sincerely, ',
+                  4:'-The Team'
+                  }
+    #Not the best way to do this.
+    txt = ("\n{} {},\n\n {:>45}${:.2f}.\n\n {: >40}.\n\n{:>40}\n"
+         "{:>42}".format(e_frmt.get(0),name,e_frmt.get(1), newAmout,e_frmt.get(2),e_frmt.get(3),e_frmt.get(4)))
```
hosungs, line 108: This looks unnecessarily concocted--I think I heard you saying that already in class... :)

I think the intention of the `dict` use requirement with `.format(...)` was to exercise the keyword arguments with a dict expansion. No need to change this for that now.

hosungs (commented):
Left some comments. Looks like trigrams assignment is still in progress, so I'll leave this PR open. Please push new commits that completes the assignment. I'll review it then and close this PR for the next assignment.

students/brandon_nguyen/session04/trigrams_text.py
bmnguyen: Thank you.  I am trying to get the practice in now.  Still tough.  No wonder dev do not like it.

students/brandon_nguyen/session04/mailroom2.py
bmnguyen: I will get back into the mailroom2 when I am done with triagrams.  I am still struggling thru it.

hosungs:

PR #58, opened 2018-11-01
Session 04 PR

funnyletter:


students/rachel_s/session04/kata14.py
```diff
@@ -1,12 +1,25 @@
...
+        found_start = False
+        for line in f.readlines():
+            if not found_start:
+                if "***" in line:
+                    found_start = True
+            else:
+                if "***" in line:
+                    break
+                line = re.findall(r"[\w']+|[.,!?;]", line.lower().strip()
+                .replace("\n", " "))
```
hosungs, line 20: Great to see your regex skill already!

students/rachel_s/session04/kata14.py
```diff
@@ -24,11 +37,51 @@ def build_trigrams(words):
         if not pair in trigrams:
             trigrams[pair] = [follower]
         else:
-            trigrams[pair] += [follower]
-
+            trigrams[pair].append(follower)
```
hosungs, line 40: Minor: Lines 37-40 can be simply `trigrams.setdefault(pair, []).append(follower)`.

Or even better would be defining `trigrams = defaultdict(list)` (`from collections import defaultdict` at the top of the script), and then just say `trigrams[pair].append(follower)`.

students/rachel_s/session04/kata14.py
```diff
@@ -24,11 +37,51 @@ def build_trigrams(words):
...
+def build_text(word_pairs):
+    output = []
+    for i in range(10):
+        starter = start_sentence()
+        make_sentence(output, starter)
+    
+    return output
+
+def start_sentence():
+    starters = list(filter(lambda tup: tup[0] == ".",  word_pairs.keys()))
```
hosungs, line 53: Great to see functional `filter(...)` and `lambda`!

Now you learned list comprehension, you may use it for this kind of occasion in the future... :)

I'm a little unsure, though, about the filtering condition (`lambda tup: tup[0] == "."`). Does this mean that some of your pairs might have `"."` as its first part? I didn't check your regex carefully, so I suppose it might be possible, but then I suppose the desired intention was to get rid of all the punctuation, not retain them. Well, there's no hard requirement for that, so I suppose this is fine.

students/rachel_s/session04/kata14.py
```diff
@@ -24,11 +37,51 @@ def build_trigrams(words):
...
+    return random.choice(starters)
+
+
+def make_sentence(output, pair):
+    if pair[1] == ".":
+        output.append(pair[1])
+    else:
+        output.append(pair[1])
+        new_pair = (pair[1], random.choice(word_pairs.get(pair)))
+        make_sentence(output, new_pair)
```
hosungs, line 63: Interesting use of recursion, and also valid, but hopefully you don't get the stack overflow exception, as the length of the generated output (which will be about equal to the recursion depth) grows... It's best to use a loop in such a case like this, instead of recursion.

students/rachel_s/session04/mailroom_with_dicts.py
```diff
@@ -0,0 +1,93 @@
...
+import csv
+import os
+import datetime
+
+# Import donor data
+
+donor_info = {}
+
+with open('mailroom/donors.csv') as csv_file:
+    csv_reader = csv.reader(csv_file, delimiter=',')
```
hosungs, line 12: Great to see your csv library usage!

students/rachel_s/session04/mailroom_with_dicts.py
```diff
@@ -0,0 +1,93 @@
...
+    for line in csv_reader:
+        if not line[0].lower() in donor_info:
+            donor_info[line[0].lower()] = [int(line[1])]
+        else:
+            donor_info[line[0].lower()].append(int(line[1]))
```
hosungs, line 17: Minor again: Consider replacing lines 14-17 to `donor_info.setdefault(lines[0].lower(), []).append(int(line[1]))`, or better yet use `defaultdict`.

students/rachel_s/session04/mailroom_with_dicts.py
```diff
@@ -0,0 +1,93 @@
...
+        add_donor(donor_name, donation)
+        print(generate_letter(donor_name))
+
+
+def add_donor(name, amount):
+    name = name.lower()
+    if name in donor_info:
+        donor_info[name].append(amount)
+    else:
+        donor_info[name] = [amount]
```
hosungs, line 47: Same minor comment: Consider using `.setdefault(...)`, or `defaultdict`.

students/rachel_s/session04/mailroom_with_dicts.py
```diff
@@ -0,0 +1,93 @@
...
+        response = input("Please select 1, 2, 3, or 4.\n> ")
+
+
+# Send a thank you
+def send_thank_you():
+    donor_name = input("What is the donor's full name?\n> ").lower()
+    if donor_name == 'list':
+        for i in donor_info.keys():
+            print(i.title())
+        user_menu()
```
hosungs, line 35: Uhh... You are using undesired recursion for the main menu loop. It's best not to use recursion for that purpose. It's confusing, like it's not obvious how the loop works, when some functions call `user_menu()` at the end, but other functions don't. I thought you used the desired while loop in your mailroom part 1, and I'm wondering why you changed that structure... Please see [this](https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom_tutorial.html#controlling-main-program-flow) regarding the recommended main menu loop control structure.

hosungs (commented):
Looks good, though I'd prefer to see revised main menu loop control structure in your mail room part 2. Merging/closing this PR for the next assignment.

PR #59, opened 2018-11-01
assignment mailroom2

ikemerrixs:


students/merrix/session02/fizzbuzz.py
```diff
@@ -0,0 +1,9 @@
+for num in range(1, 101):
+    if num % 15 is 0:
```
hosungs, line 2: Best not to use `is` when comparing numbers (or almost anything other than `x is None`).

students/merrix/session03/mymailroom.py
```diff
@@ -0,0 +1,101 @@
...
+          "4 - To Exit",
+          ">>> "))
+
+
+def view_donor_db():
+    print (donor_db.keys())
+    choose = input("enter a name from the list or a new name to donate: ").title()
+    if choose not in donor_db.keys():
+        add_donor_db(choose)
+    elif choose in donor_db.keys():
```
hosungs, line 28: Minor: No need to have `elif` with the negation of the `if` condition, because `else` alone implies the negation of the `if` condition anyway.

students/merrix/session03/mymailroom.py
```diff
@@ -0,0 +1,101 @@
...
+    new_donor = input("Do you want to add %s to the donation list? yes or no " % choose).title()
+    if new_donor == "Yes":
+        amount = input("Enter an amount to donate for %s " % choose)
+        donor_db.update({new_donor:amount})
+        print (donor_db)
+        print ('''
+            Thank you %s for your donation of $%i to this organizat1ion. We really appreciate your generousity and hope to see you again at our upcoming events.
+            Sincerely Yours
+            James Dot
+            ''' % (new_donor, int(amount)))
```
hosungs, line 52: There are duplicates in your code, here in lines 47-52 and above in lines 31-37. I understand you had to handle the case of a new donor separately, because you have to initialize the new donor entry. However, once that's initialized, printing the letter is no different from the case of an existing donor, so this duplicate could have been avoided. It's OK for now, but please keep in mind that avoiding unnecessary duplicated code is very important in programming.

students/merrix/session03/mymailroom.py
```diff
@@ -0,0 +1,101 @@
...
+    for name, donations in donor_db.items():
+        num_of_gift = len(donations)
+        total_given = int(sum(donations))
+        average_gift = int(total_given/num_of_gift)
+        print ('{0:<15}   {1:^10}     {2:^6}     {3:8}'.format(name, total_given, num_of_gift, average_gift))
+        
+
+def send_letters():
+    for donor_name in donor_db:
+        letter = open('{}.txt'.format(donor_name), "w+")
```
hosungs, line 69: It's best to use `with open(...) as letter:`, so that the opened file is automatically closed. You are not using `with`, and not explicitly closing the written files, so you are risking losing your written content, if program crashes unexpectedly.

students/merrix/session04/mymailroom.py
```diff
@@ -0,0 +1,101 @@
...
+            send_letters()
+        elif response == "4":
+            exit_program()
+        else:
+            print("Not a valid option!")
+
+
+if __name__ == "__main__":
+    # don't forget this block to guard against your code running automatically if this module is imported
+    main()
```
hosungs, line 101: This file looks identical to your `session03/mymailroom.py`. I guess you ended up catching up directly to session 4, so didn't actually implement the mailroom in session 3 ways. Not sure how it'll be graded, though, in our course management system (Canvas). I'll leave that up to our TA (@Saravji -- Please let me know if you'd like to discuss this more).

hosungs (commented):
Left some comments, will merge/close this PR for the next assignment.

PR #61, opened 2018-11-04
Laud Tetteh: Session 04

laudtetteh:
MailRoom 2 for Session04 is not yet included here because because I'm yet to complete and submit MailRoom1 for Session03. Working on all my missing assignments at the moment.


students/laud/session04/kata_fourteen.py
```diff
@@ -0,0 +1,63 @@
...
+
+# Build key-value pairs
+def build_trigram(words):
+    trigrams = {}
+    for i in range(len(words)-2):
+        pair = tuple(words[i:i + 2])
+        follower = words[i + 2]
+        if pair not in trigrams:
+           trigrams[pair] = []
+        trigrams[pair].append(follower)
```
hosungs, line 33: Minor: Lines 31-33 can be simplified to `trigrams.setdefault(pair, []).append(follower)`, or even define `trigrams = collections.defaultdict(list)` (after `import collections` at the top of this source file.

students/laud/session04/kata_fourteen.py
```diff
@@ -0,0 +1,63 @@
...
+    return trigrams
+
+# Generate new text
+def build_text(word_pairs):
+    word_list = []
+    starting_pair = list(random.choice(list(word_pairs)))
+    word_list.extend(starting_pair)
+    for x in range(len(word_list)):
```
hosungs, line 41: Does this `for` loop work? At the time of the loop's beginning, `len(word_list)` is fixed to 2, and I guess this loop will just iterate twice, not any more? Basically your `word_list` is modified inside the loop, and you are using its length as the range in the loop header, which is very undesirable, if not incorrect. I think you might want to fix the number of iterations of this `for` loop somehow, like just hard-coding a fixed number or using a passed parameter?

hosungs (changes requested):
I'm not sure why this PR contains changes on other students' directories. With those in this PR, we cannot merge this PR at all. We need to fix this, but I'm not getting any quick/easy fix for this issue. We may have to plan on fixing this together online through Zoom screensharing. If we can't do that before next class, we'll have to do that in class coming Tuesday. Just FYI.

students/laud/session04/kata_fourteen.py
laudtetteh: @hosungs Thanks for the feedback. I'll attempt to fix these PRs, but yes, we can take a look at them in class. Remind me, though, isn't it this Tuesday that you're going to be out of town?

students/laud/session04/kata_fourteen.py
laudtetteh: @hosungs Thanks. I'll take another look.

hosungs: Yes, I'm out of town the whole week this week, but I'll be teaching the class on Wednesday through Zoom, so we can handle this on Wednesday during the class time.

laudtetteh: @hosungs Sounds good. Thank you.

students/laud/session04/kata_fourteen.py
laudtetteh: @hosungs I just remembered how this happened! I think I accidentally ran `git pull upstream master` before committing my stuff. I'll just go ahead and remove the stuff that showed up from the other students' folders, and then re-commit. I think that should do it.

laudtetteh: @hosungs Also, should I just go ahead and maintain this one PR for all my late assignments, instead of attempting to create a separate PR for Session03 and Session04? Seems to me that would be less messy.

laudtetteh:
Closing this PR to be replaced by a cleaner one.

students/laud/session04/kata_fourteen.py
hosungs: > @hosungs Also, should I just go ahead and maintain this one PR for all my late assignments, instead of attempting to create a separate PR for Session03 and Session04? Seems to me that would be less messy.

Yes, we should be able to do with one single PR. Thanks.

PR #62, opened 2018-11-04
Laud Tetteh: Session03

laudtetteh:
This is a separate branch just for Session03 late submissions. The first is list_lab.py. Working on the other 3 assignments in Session03 at the moment. This branch will be merged back into my personal Master. Then another PR will be created. Please forgive my messy attempt at organizing my mess!


hosungs (changes requested):
It's great to see that you learned how to use non-master branches, but there's no need for that, if you are not careful with correctly separating your work. Besides, this change includes your kata fourteen work as well, which is also included in your other PR. This is not ideal. Also, this branch includes changes in other students' directories, which shouldn't have happened at all. As I mentioned on the other PR, we should plan on fixing this all in person or online interactively. Or please try to fix this yourself and see if this can be fixed.

laudtetteh:
@hosungs I think I'm just going to clean things up and use only Master. However, please note that I'll be using one PR for both Session03 and Session04 (I'll named that PR accordingly). I've also removed all the stray changes from other students' directories. And I'll alert you when I'm done cleaning up, and ready for review. Proceeding to remove both of my current PRs right now to be replaced with one, clean one. Thanks again for your patience. I'll get better at this!

PR #63, opened 2018-11-05
Commits for session05

motians:
It was a little unclear as to what parts of the exception and comprehension exercises should be submitted. I added exception handling to the mailroom but comprehensions did not seem to fit in the program.


students/shawn_degraw/session05/mailroomPart3.py
```diff
@@ -0,0 +1,156 @@
...
+        filename = "letters\\" + donorname.replace(' ', '_') + ".txt"
+        # create a new dictionary just for the format mailroom part 2
+        formatdict = {"name": donorname, "totaldonation": sum(donor_db[donorname])}
+        try:
+            with open(filename, 'w') as outfile:
+                outfile.write(GENERAL_DONATION_LETTER.format(**formatdict))
+        except FileNotFoundError:
+            try:
+                makedirs("letters")
+                sendletters()
```
hosungs, line 114: Interesting nested try-except blocks, with another interesting recursion... I think it's best to avoid this recursion all together, but I understand it may feel nontrivial to do so. You may want to separate the `letters` subdirectory part separately, so that this doesn't have to be nested like this. With this nesting, I'm not sure what will happen if some exception gets thrown in a lower level recursive call...

hosungs (commented):
Saw your message, and it appears that there are not really many places where comprehension can be applied... So I think this should be good enough.

I left a comment about the nested try-except blocks and the unusual/undesired recursion. I'll leave it up to you whether to change it or not, so just merging/closing this PR as completed.

PR #64, opened 2018-11-06
PR for session 5

yushusongmsazure:
Includes mailroom3 and exception_excise for session 05


students/yushus/session05/mailroom3.py
```diff
@@ -0,0 +1,120 @@
...
+        if name.lower() == 'list':
+            print_donor()
+        elif name.lower() not in donors: # New donor
+            donor_db[name]=[]
+            break
+        else: # Existing donor; find the correct donor name from DB
+            for donor_name in donor_db:
+                if donor_name.lower() == name.lower():
+                    name = donor_name
+            break
```
hosungs, line 62: Hmm, I must have overlooked this part. Once you have a dict for `donor_db`, this `for` loop is not really needed. Remember that the only reason we had this loop was because with a list for `donor_db`, we couldn't say `donor_db[name]`, so we had to go through all items in the list and find out the matching one. With a dict, there's no need to go through all items.

This is related to the earlier assignment, not this one, so this change is not required, but I'd say it'd be good to do that anyway. I'll leave it up to you.

students/yushus/session05/mailroom3.py
```diff
@@ -0,0 +1,120 @@
...
+            amount = float(input("How much do you want to donate? "))
+            if amount <= sys.maxsize and amount > 0:
+                donor_db[name].append(amount)
+                print_email(name, amount)
+            else:
+                raise ValueError()
+        except ValueError:
+            print("Please enter a valid donation!")
+        else:
+            break
```
hosungs, line 76: Good `try`-`except`-`else` usage!

hosungs (commented):
Left some comments, and will merge/close this PR, as it looks complete for the week (though it'd be great to see more list comprehension in your new code, but I can't think of anything else myself, so...)

students/yushus/session05/mailroom3.py
yushusongmsazure: The thing I am trying to achieve is to ignore cases when user provides the donor name. Since dict is case sensitive for the keys, I have to find the "original" key by loop through the dict.

hosungs: I see. It's more like a data/input sanitization issue. In that case, I suppose it'd be still better to avoid this kind of sequential search with a dict, by for example, standardizing the keys (like `str.title()`?). If you have to do sequential search like this with a dict, there's really no point using a dict at all.

PR #65, opened 2018-11-06
Session 05 Mailroom 3 and except_exercise Ying Guo

guoguoying2013:


students/YingGuo/session05/mailroom3.py
```diff
@@ -0,0 +1,67 @@
...
+
+            elif user_input.lower() == "quit":
+                break
+            
+            else:
+                amount1 = int(input("how much would you like to donat?"))
+                data[user_input.capitalize()] += [amount1]
+                print("Thank you {} for your donation {}!".format(user_input.capitalize(), amount1))
+                return data
+        except ValueError:
```
hosungs, line 26: I think this will work, but it looks to me the scope of the try-except block is unnecessarily wider than necessary. I believe the `ValueError` is for line 22 (`int(input(...))`) only, and it's best to narrow down the try-except scope to the needed code only.

students/YingGuo/session05/mailroom3.py
```diff
@@ -0,0 +1,67 @@
...
+            print("The value entered is invalid")
+
+
+#Creat a report function
+def report(data):
+    """this function will print a report of donation list with data, total donation, donation times, donation average"""
+    report_list = [(x, sum(data[x]), len(data[x]), sum(data[x])/len(data[x])) for x in data.keys()]
```
hosungs, line 33: Great comprehension you come up with!

students/YingGuo/session05/mailroom3.py
```diff
@@ -0,0 +1,67 @@
...
+
+if __name__ == "__main__":
+    while True:
+        try:
+            user_input = int(input("1. send a thank you to a donor\n2. report\n3. send letter to all\n4. quit"))
+            if user_input == 4:
+                break
+            dct_opt = {1:thank_you, 2:report, 3:all_letter}
+            dct_opt.get(user_input)(data)
+        except TypeError:
```
hosungs, line 66: Is this `TypeError`? I thought it's `ValueError` when `int(input(...))` fails with a nonnumeric input?

Also, what about `KeyError` when unavailable menu number (e.g., 0 or 5 or 6 or ...) is entered?

hosungs (commented):
Left a few comments. I'd like the main menu loop comment to be addressed (`ValueError` and `KeyError`), but I won't require that for completion of this assignment, so I'll merge/close this PR as is. Feel free to make changes and push more commits.

guoguoying2013:
dict.get() doesn't look like raise error if out of index. In this case, the
TypeError works.
I added ValueError in the new PR. Thank you for your comments.

PR #66, opened 2018-11-06
Laud Tetteh: Session03 & Session04

laudtetteh:
This new PR contains all 4 assignments from Session03. I attempted to keep this one free of stray files from other students' directories. Please do let me know if I need to do more cleaning up. Thanks, Hosung!


hosungs (commented):
Yes, one PR for all assignments in a single week is exactly what's desired, so you did correctly with this PR.

Unfortunately, you removed other students' file, not reverting your changes on those other students' files. Removing (deleting) other students' files is not the same as reverting your changes on those other students' files. This has become more complicated than desired, so we'll need to fix this together in class tomorrow...

Also, your mailroom shows quite undesirable recursive pattern, which makes it very hard to understand your program flow, and I'm not sure how to tackle that... We might need to go over that aspect together in class...

Leaving this PR open as such...

students/laud/session03/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+        print()
+        get_main_menu()
+    else:
+        if   choice == 1: send_thank_you()
+        elif choice == 2: create_report(donations)
+        elif choice == 3: quit_program()
+        else:
+            print("Oops! Please make a valid selection.")
+            print()
+            get_main_menu()
```
hosungs, line 120: Still undesirable recursion-based menu loop... I'll have to leave the validation of this recursion to TA, and have to ask you to remove all these undesirable recursions in the newer mailroom assignment parts.

students/laud/session04/kata_fourteen.py
```diff
@@ -0,0 +1,62 @@
...
+    word_list = []
+    starting_pair = list(random.choice(list(word_pairs)))
+    word_list.extend(starting_pair)
+    for x in range(int(limit)):
+        current_key = tuple(word_list[-2:])
+        if current_key in word_pairs:
+            next_word = random.choice(list(word_pairs[current_key]))
+            word_list.append(next_word)
+        else:
+            continue
```
hosungs, line 45: This is effectively the same as `break`, because continuing the loop will keep seeing the same last 2 words as the `current_key`, which won't be in `word_pairs` again (because nothing changed), so keep hitting this `else` block repeatedly until the `range(int(limit))` is exhausted. Better just use `break`.

students/laud/session04/mailroom2.py
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env python3
+import sys
+
+""" Historical donations data """
+donations = [
+    ("Bill Gates", [653772.32, 12.17]),
+    ("Jeff Bezos", [877.33]),
+    ("Paul Allen", [663.23, 43.87, 1.32]),
+    ("Mark Zuckerberg", [1663.23, 4300.87, 10432.00]),
+]
```
hosungs, line 10: In session 4 mailroom part 2, you are required to use a Python dictionary, not a list, so this needs to be changed.

In fact, browsing this source overall, this looks like the same as your session 3 mailroom part 1 source code. So it looks like you really haven't tried session 4 mailroom part 2 yet.

Please do try session 4 mailroom part 2 with a dictionary and file writing feature implemented.

hosungs (commented):
Session 3 mailroom part 1 may be considered OK, though there's still some undesirable recursion-based menu loop handling.

Session 4 Kata Fourteen (Trigrams) seems OK.

Your session 4 mailroom part 2 source code is same as your session 3 mailroom part 1, so session 4 mailroom part 2 is not completed in this PR.

I'll close this PR for completed session 3 mailroom part 1 and session 4 trigrams. Please keep working on session 4 mailroom part 2 and all the delayed assignments (session 5 mailroom part 3 and session 6 mailroom part 4) as early as possible!

PR #67, opened 2018-11-06
Added Lesson5 assignments - Mailroom-Part3 and exception exercise

nallarun10:
PR_Lesson05

### DIFF
--- a/students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
+++ b/students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
@@ -1,0 +1,99 @@
+#!usr/bin/env python
+
+donors_list = [('Aron King', [1, 2]), ('Becky Patty', [3000, 1000]),
+               ('Charlie Lee', [666, 222]), ('Thomas Dave', [202, 90, 450]), ('Nancy Crow', [100, 200, 300])]
+# input from Mairoom_Part1. Convert into a dict
+
+donors_dict = {x: y for x, y in donors_list}
+
+def thanks_email():
+    while True:
+        name_input = input("Enter full name of a donor>>>").title()
+        if name_input == 'List': print(list(donors_dict))
+        else:
+            if name_input in donors_dict:
+                d = {}
+                d['name'] = name_input
+                while True:
+                    try:
+                        donation_amount = int(input('Enter donation amount: '))
+                    except ValueError:
+                        print ('Please enter current value, entry should be an integer.')
+                    else:
+                        donors_dict[name_input].append(donation_amount)
+                        d['donation'] = donation_amount
+                        print("\nDear {name},\n\tThank you for generous donation of ${donation:.2f} to ZYX charity\n"
+                          "Sincerely,\nZYX Team".format(**d))
+                        return
+            else:
+                while True:
+                    d = {}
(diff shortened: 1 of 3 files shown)


students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
+#!usr/bin/env python
+
+donors_list = [('Aron King', [1, 2]), ('Becky Patty', [3000, 1000]),
+               ('Charlie Lee', [666, 222]), ('Thomas Dave', [202, 90, 450]), ('Nancy Crow', [100, 200, 300])]
+# input from Mairoom_Part1. Convert into a dict
+
+donors_dict = {x: y for x, y in donors_list}
```
hosungs, line 7: I must have overlooked this initialization code. No need to create a separate list just for dict initialization. Can do the initialization directly:
```
donors_dict = { 'Aron King': [1, 2], 'Becky Patty': [3000, 1000], ... }
```
I might have already mentioned this... I'll remember that and stop making any more repeated comments... :)

students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
...
+                        donation_amount = int(input('Enter donation amount: '))
+                    except ValueError:
+                        print ('Please enter a current value, entry should be an integer.')
+                    else:
+                        donors_dict[name_input] = [donation_amount]
+                        d['donation'] = donation_amount
+                        #dict_format(seq,name_input)
+                        print("\nDear {name},\n\tThank you for generous donation of ${donation:.2f} to ZYX charity\n"
+                          "Sincerely,\nZYX Team".format(**d))
+                        return
```
hosungs, line 42: Almost identical code (lines 17-27 and lines 29-42), which is not desirable at all. I think I might have already mentioned this as well... You could have used `donors_dict.setdefault(name_input, []).append(donation_amount)` to unify the only different part (line 23 and line 37).

students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
...
+        try:
+            response = input("MAIN MENU\nDonors and donations made to XYZ organisation.\nPlease select one of the option"
+                         " below to proceed.\n1.Send a Thank you message to the donor\n2.Create a report\n3.Send "
+                            "letters to all donors\n4. Quit\n>>>")
+
+            if not response in dict_list:
+                print ('Invalid input.\nPlease select correct option')
+            else:
+                dict_list.get(response)()
+        except ValueError:
```
hosungs, line 93: Hmm, certainly not `ValueError`, because you are not converting the `input(...)` result to a number. However, it won't be `KeyError` either, because you used `dict_list.get(...)` (which doesn't throw but return `None` if the passed key is not present in the dict), not `dict_list[...]` (which does thrown `KeyError` in the same case). Did you test your code for those cases?

hosungs (commented):
Left some comments... Unsure if I can declare this as complete, but I'll merge/close this PR anyway, as it includes all submissions anyway. Feel free to add more commits and possibly create another PR if you wish.

PR #68, opened 2018-11-07
Session 05 pull request

funnyletter:


students/rachel_s/session05/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+
+def user_menu():
+    menu = {1: send_thank_you, 2: create_report, 3: all_donor_letters, 
+    4: exit_program}
+    response = input("\n\nSelect an option:\n  1  Send a thank you to a "
+    "single donor\n  2  Create a report\n  3  Send letters to all donors"
+    "\n  4  Quit\n\n> ")
+    try:
+        menu.get(int(response))()
+    except(KeyError, ValueError):
```
hosungs, line 32: Good to see you are catching multiple types of exceptions!

students/rachel_s/session05/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+def all_donor_letters():
+    user_dir = input("What directory do you want to write to?\n\n>  ")
+    dest = "{user_dir}/{date}_{name}.txt"
+    date = datetime.datetime.today().strftime("%Y-%m-%d")
+    if not os.path.isdir(user_dir):
+        os.makedirs(user_dir)
+    for name in donor_info.keys():
+        path = dest.format(user_dir=user_dir, name=name, date=date)
+        with open(path, "w") as f:
+            f.write(generate_letter(name))
```
hosungs, line 82: Maybe a good place to add some try-except for file writing failures (e.g., disk full, no permission, ...)...?

students/rachel_s/session05/mailroom.py
```diff
@@ -0,0 +1,105 @@
...
+import sys
+from collections import defaultdict
+
+# Import donor data
+def data_import(filename):
+    donor_info = defaultdict(list)
+    with open(filename) as csv_file:
+        csv_reader = csv.reader(csv_file, delimiter=',')
+        for line in csv_reader:
+            donor_info[line[0].lower()].append(int(line[1]))
```
hosungs: Looking for where you applied any list/dict/set comprehension, and not sure if I found any.

I thought this might be a good place to apply list comprehension, but maybe not... Since the assignment requires some list comprehension, could you try adding some? I know it feels like making up, but it's for practice...

hosungs (commented):
Left a few comments... Not sure if I should merge this PR or leave it open... I'll leave it open for a while...

students/rachel_s/session06/test_mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+    os.remove('thisisatest/' + today + '_one.txt')
+    assert os.path.isfile('thisisatest/' + today + '_two.txt') == True
+    os.remove('thisisatest/' + today + '_two.txt')
+    assert os.path.isfile('thisisatest/' + today + '_three.txt') == True
+    os.remove('thisisatest/' + today + '_three.txt')
+    assert os.path.isfile('thisisatest/' + today + '_four.txt') == True
+    os.remove('thisisatest/' + today + '_four.txt')
+    assert os.path.isfile('thisisatest/' + today + '_decimal.txt') == True
+    os.remove('thisisatest/' + today + '_decimal.txt')
+    os.rmdir('thisisatest')
```
hosungs, line 66: Great job that you are checking that files are generated! More great is that you cleaned up the generated files/dir so that any next run wouldn't be affected by the files/dir that were generated in a previous run!

students/rachel_s/session06/test_mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+    out = create_report(db)
+    assert len(out) == 7
+    assert out[0] == ('\n\nDonor Name           |   Total Given |    Num Gifts'
+    ' |      Avg Gift')
+    assert out[2] == ('One                  | $        1.00 |            1 |'
+    ' $        1.00')
+    assert out[4] == ('Three                | $        6.00 |            3 |'
+    ' $        2.00')
+    assert out[6] == ('Decimal              | $        4.94 |            2 |'
+    ' $        2.47')
```
hosungs, line 78: Minor: Testing on printed strings is very fragile (can fail even for a single space difference), so it's rarely done. Output validation is not easy, and probably not in the scope of unit testing. Still I do see some value in doing this and understand your intention. Maybe parsing the printed string back to some data structure and comparing those parsed values might be a better approach, but certainly not expected in this assignment.

hosungs (approved):
Good job! This PR now contains both assignments 5 and 6, and both look good. Will close this now for the upcoming assignment. Thanks!

PR #69, opened 2018-11-07
updated mailroom3

guoguoying2013:
instructor's comment:

"if __name__ == "__main__":
+    while True:
+        try:
+            user_input = int(input("1. send a thank you to a donor\n2. report\n3. send letter to all\n4. quit"))
+            if user_input == 4:
+                break
+            dct_opt = {1:thank_you, 2:report, 3:all_letter}
+            dct_opt.get(user_input)(data)
+        except TypeError:
Is this TypeError? I thought it's ValueError when int(input(...)) fails with a nonnumeric input?

Also, what about KeyError when unavailable menu number (e.g., 0 or 5 or 6 or ...) is entered?"

Reply:

dict.get() doesn't raise error if out of index so I used TypeError, it works in this case.
I just added ValueError in this new PR.


hosungs (approved):
Looks good

PR #70, opened 2018-11-07
added Mailroom_Part3

nallarun10:


students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
+#!usr/bin/env python
+
+donors_list = [('Aron King', [1, 2]), ('Becky Patty', [3000, 1000]),
+               ('Charlie Lee', [666, 222]), ('Thomas Dave', [202, 90, 450]), ('Nancy Crow', [100, 200, 300])]
+# input from Mairoom_Part1. Convert into a dict
+
+donors_dict = {x: y for x, y in donors_list}
```
hosungs: I may be repeating myself again, but no need to initialize `donors_dict` indirectly like this. You can initialize any dict directly by saying `d = { 'k1': 'v1', 'k2': 'v2', ...}`.

students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
...
+                        donation_amount = int(input('Enter donation amount: '))
+                    except ValueError:
+                        print ('Please enter a current value, entry should be an integer.')
+                    else:
+                        donors_dict[name_input] = [donation_amount]
+                        d['donation'] = donation_amount
+                        #dict_format(seq,name_input)
+                        print("\nDear {name},\n\tThank you for generous donation of ${donation:.2f} to ZYX charity\n"
+                          "Sincerely,\nZYX Team".format(**d))
+                        return
```
hosungs: Your liens 15-27 and 29-42 are very similar with lot of duplicate code. They should be refactored using `dict.setdefault()` or `defaultdict` or restructuring the `if`-`else` code, to avoid code duplication.

students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
...
+
+def sort_key(sorting):
+    return sorting[1]
+
+def create_report():
+
+    list_donors = [[a, sum(donors_dict[a]), len(donors_dict[a]), sum(donors_dict[a])/len(donors_dict[a])] for
+                   a in list(donors_dict)]
```
hosungs: Good comprehension! One minor thing to comment is that it'd be better if each element in the list is a tuple (`(a, sum(donors_dict[a]), len(donors_dict[a]), sum(donors_dict[a])/len(donors_dict[a]))`) instead of a list(`[a, sum(donors_dict[a]), len(donors_dict[a]), sum(donors_dict[a])/len(donors_dict[a])]`). Functionally this doesn't change anything, but still a good practice to use tuples for items, not lists.

students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
...
+    list_sorting = (sorted(list_donors, key=sort_key, reverse=True))
+    print(list_sorting); print("\nList of donors and Contribution to XYZ Charity\n")
+    print("{:<20}\t\t{:<10}\t\t{:<10}\t\t{:<10}".format('Donors-name', 'Total', 'Num gifts', 'Average gift'))
```
hosungs: I suppose I might have overlooked this earlier, but it's not recommended to use `\t` in string formatting, because you wouldn't know how many spaces each tab will be. That's why there are extensive formatting syntax, specifying various widths.

students/arun_nalla/session05/Lesson5_Mailroom_Part3.py
```diff
@@ -0,0 +1,99 @@
...
+    while True:
+        try:
+            response = input("MAIN MENU\nDonors and donations made to XYZ organisation.\nPlease select one of the option"
+                         " below to proceed.\n1.Send a Thank you message to the donor\n2.Create a report\n3.Send "
+                            "letters to all donors\n4. Quit\n>>>")
+
+            if not response in dict_list:
+                print ('Invalid input.\nPlease select correct option')
+            else:
+                dict_list.get(response)()
```
hosungs: This `if`-`else` (asking permission) could have been another `try`-`except` (with `KeyError`, asking forgiveness, which is considered better than asking permission).

students/arun_nalla/session05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,77 @@
...
+            while True:
+                try:
+                    donation_amount = int(input('Enter donation amount: '))
+                    donors_dict.setdefault(name_input, []).append(donation_amount)
+                    d = {}
+                    d['name'] = name_input
+                    d['donation'] = donation_amount
+                    print("\nDear {name},\n\tThank you for generous donation of ${donation:.2f} to ZYX charity\n"
+                      "Sincerely,\nZYX Team".format(**d))
+                    return
```
hosungs, line 21: Uh... OK, please ignore my comments above, because you definitely addressed those in this new source file...

students/arun_nalla/session05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,77 @@
...
+                    print ('Please enter a value, entry should be an integer.')
+
+
+def sort_key(sorting):
+    return sorting[1]
+
+def create_report():
+
+    list_donors = [[a, sum(donors_dict[a]), len(donors_dict[a]), sum(donors_dict[a])/len(donors_dict[a])] for
+                   a in list(donors_dict)]
```
hosungs, line 34: Well, this one still uses lists as items, and they better be tuples as I mentioned above.

students/arun_nalla/session05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,77 @@
...
+    list_sorting = (sorted(list_donors, key=sort_key, reverse=True))
+    print("\nList of donors and Contribution to XYZ Charity\n")
+    print("{:<20}\t\t{:<10}\t\t{:<10}\t\t{:<10}".format('Donors-name', 'Total', 'Num gifts', 'Average gift'))
```
hosungs: Avoid using `\t` in formatting strings, as mentioned above.

hosungs (commented):
Left some comments. No significant issues, so I was going to merge this PR, but there are some merge conflicts (with some files on some other students' directories, which shouldn't have happened in the first place), so we'll need to leave this PR open and fix this issue later in class...

PR #71, opened 2018-11-07
week5 HW

aminavi:


students/aminavi/session05/mailroom3.py
```diff
@@ -0,0 +1,87 @@
...
+    while True:
+        try:
+            response = int(input(prompt))
+        except ValueError:
+            print("\n Input must be between 1 and 4 (integer), try again. \n")
+            continue
+        if response in function_list.keys():
+            function_list.get(response)()
+        else:
+            print("Not a valid option!")
```
hosungs, line 84: This `if`-`else` (asking permission) could have been just another `try`-`except` (asking forgiveness, which is considered better than asking permission).

students/aminavi/test.py
```diff
@@ -1,93 +1,94 @@
...
+    path = input("please enter the path ")
+    if path != '':
+        path = path + "\\"
+    for k,v in donor_list.items():
+        with open("{}{}.txt".format(path, k), "w") as letter:
+            letter.write("Dear {},\n\n".format(k))
+            letter.write("\tThank you for your generous donation of ${}\n\n".format(sum(v)))
+            letter.write("\tIt will be put to very good use.\n\n")
+            letter.write("\t\tSincerely,\n")
+            letter.write("\t\t  - Team\n")
```
hosungs, line 48: Would have been better if some file-related errors (e.g., `PermissionError` that's raised when trying to write to a location where you don't have access) were handled using `try`-`except`...

hosungs (approved):
Left a few comments for more desired addition, but looks good overall.

PR #72, opened 2018-11-07
 Session05 assignment. Minor fix on Session04 per PR feedback

ejackie:


students/ejackie/session05/mailroom3.py
```diff
@@ -0,0 +1,113 @@
...
+            if dispatch_dict[response]() == "exit program":
+                break
+        except KeyError:
+            print("Please choose a correct number!\n")
+
+
+def send_a_thank_you():
+    name = input("Please enter the full name of the donor: ")
+    if name == "list":
+        [print(" ".join(key)) for key in donor_dict]
```
hosungs, line 41: I guess you tried to meet the comprehension requirement of the assignment, and I know there doesn't seem to be good places where we can apply list comprehension, so you did this eventually, but this (just printing something using comprehension, not creating a list) is not considered a good practice. Besides, `key` here will be only one value, not a list, so there won't be any good effect coming out of this. Actually, since `key` will be a string, and a string will be broken into characters as a sequence in this case (because `str.join()` takes a sequence), the printed result will be interleaved with blank spaces, which is not desired.

I'd do something like this?
```
print(" ".join(key for key in donor_dict))
```
In fact, we don't even need this, because a Python dict itself will iterate over keys if it's passed to a function as a sequence. So the above statement can be just written as `print(" ".join(donor_dict))`, and only keys will be printed. So again, back to where we were, where we didn't seem to find good candidates to convert to list/dict/set comprehensions...

students/ejackie/session05/mailroom3.py
```diff
@@ -0,0 +1,113 @@
...
+        print(send_email(name, amount))
+    else:
+        print("Adding new donor entry to the db")
+        try:
+            amount = float(input("Please enter the amount of donation: "))
+        except ValueError:
+            print("The amount entered is invalid! Please try again later.\n")
+            return None
+        donor_dict[tuple(name.split())] = [amount]
+        print(send_email(name, amount))
```
hosungs, line 60: Your lines 44-51 and 53-60 are almost identical (except the print message), so it better be refactored by using `dict.setdefault()` or `defaultdict` or restructuring the `if`-`else` statement block...

hosungs (commented):
Left a few comments, no significant issues, so will merge the PR.

PR #74, opened 2018-11-08
Session4 Assignments

bmnguyen:
To upload mailroom3


students/brandon_nguyen/session05/mailroom3.py
```diff
@@ -0,0 +1,168 @@
...
+def sort_sum():
+    # creating a new list with computed value for easy printing.
+    """
+    This function returned a sorted list of by on order amount3
+    """
+    dblist = []
+    dbSum = []
+    # quick way to convert back to list to reuse existing code.
+    # list comprehension
+    [dblist.append((i, j)) for i, j in donor_db.items()]
```
hosungs, line 77: Not the desired usage of list comprehension. No need to separately initialize `dblist` in line 73, but just write as follows:
```
dblist = [(i, j) for i, j in donor_db.items()]
```
Still, not sure why you'd want to convert the dict to a list of pairs? In fact, an easier code for that is actually: `dblist = list(donor_db.items())`.

students/brandon_nguyen/session05/mailroom3.py
```diff
@@ -0,0 +1,168 @@
...
+    [dbSum.append((name, [round(sum(donation), 2)], [len(donation)],
+     [round(sum(donation)/len(donation), 2)]),) for name, donation in dblist]
```
hosungs, line 79: OK, another undesirable coding style with list comprehension. And you could have used just only 1 comprehension, without having to create `dblist` at all, like as follows:
```
dbSum = [(name, round(sum(donation), 2), len(donation),
     round(sum(donation)/len(donation), 2)) for name, donation in dblist.items()]
```
Note that I had to remove seemingly unnecessary list brackets around `round(...)`, `len(...)`, and so on, to simplify. Not sure why they were needed. It's best to avoid unnecessary compound/nested structures as mush as possible.

students/brandon_nguyen/session05/mailroom3.py
```diff
@@ -0,0 +1,168 @@
...
+    return sorted(dbSum, key=lambda donor: donor[1], reverse=True)
+
+
+def list_donors():
+    """
+    Basic Name of donor return in Order by First Name
+    """
+    lst = []
+    print()
+    [lst.append(person) for person in donor_db]  # list comprehension converted
```
hosungs, line 89: Another undesirable list comprehension style. Just use `lst = [person for person in donor_db]`. Better yet, just use `lst = list(donor_db.keys())` if you just need to get the list of keys only.

students/brandon_nguyen/session05/mailroom3.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def send_ty_all():
+    for personName in donor_db:
+        # assuming that the last donation appended to last
+        letter = email_template(personName, donor_db[personName][-1])
+        textfile = (personName.replace(" ", "_") + "_" +
+                    str(datetime.now()).replace(" ", "_")+".txt")
+        with open(textfile, 'w') as file_object:
+            file_object.write(letter)
```
hosungs, line 139: A good place for a `try`-`except` for cases when the file writing fails due to issues like `PermissionError` or `IOError`...

hosungs (commented):
Left some comments. No significant issues, so will merge/close this PR.

PR #76, opened 2018-11-09
Session 5 submits

davidarasim:


students/dave_arasim/session05/mailroom3.py
```diff
@@ -0,0 +1,251 @@
...
+# Global variables
+this_os = platform.system()
+quit_option = False
+
+# Form letter text
+form_text = ['Dear {form_name},\n',
+             'Thank you for your generous donation of ${form_amt:.2f}\n',
+             'Please consider this e-mail as record of your donation for tax deduction purposes.\n',
+             'Sincerely, Your Local Charity']
+form_text_str = ''.join(x for x in form_text) # Comprehension
```
hosungs, line 19: Not sure if you need a comprehension here at all? Why not just `''.join(form_text)`? I know there don't seem to be many places where comprehension is possible...

students/dave_arasim/session05/mailroom3.py
```diff
@@ -0,0 +1,251 @@
...
+    # Build the Donor Dictionary from a text file
+    donor_dict = defaultdict(list)
+    build_donor_dict(donor_dict)
+
+    # Build switch_dict with a dict comprehension
+    switch_options = [1, 2, 3, 4]
+    switch_functions = [send_thank_you, create_report, send_all_thanks, quit_option_true]
+    switch_dict = {option: function for option, function in zip(switch_options, switch_functions)}
+
+    while not(quit_option):
```
hosungs, line 36: Minor: `not quit_option`, not `not(quit_option)`, which looks like a function call, but `not` is not a function. When not is applied to a compound boolean expression, parentheses should be used, but after a blank space. E.g., `not (a or b)`.

students/dave_arasim/session05/mailroom3.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def build_donor_text(donor_dict):
+    '''Export the Donor Dictionary data back to a text file'''
+    with open('donor_dict.txt', 'w') as dictfile:
+        for this_name, this_amts in donor_dict.items():
+            amts_str = ','.join(str(x) for x in this_amts) # Comprehension
+            this_line = this_name + ':' + amts_str + '\n'
+
+            dictfile.write(this_line)
```
hosungs, line 115: Another good place for `try`-`except` in case the file writing fails due to errors like `PermissionError` or `IOError`.

students/dave_arasim/session05/mailroom3.py
```diff
@@ -0,0 +1,251 @@
...
+
+        this_tot = sum(this_amts)
+        this_file = this_name.replace(' ', '_') + '.txt'
+
+        form_dict = {'form_name': this_name, 'form_amt': this_tot}
+
+        thank_you_text = form_text_str.format(**form_dict)
+
+        with open(this_file, 'w') as thank_you_file:
+            thank_you_file.write(thank_you_text)
```
hosungs, line 205: Another good place for `try`-`catch` in case the file writing fails due to errors like `PermissionError` or `IOError`.

hosungs (commented):
Left some comments, no significant issues, so will merge/close this PR.

PR #78, opened 2018-11-13
session06 mailroompart4 and test_mailroom

motians:


students/shawn_degraw/session06/test_mailroom.py
```diff
@@ -0,0 +1,76 @@
...
+
+def test_report():
+    report = mailroompart4.createreport()
+    assert report[0] == ('\nDonor Name                | Total Given | Num Gifts | Average Gift ')
+    assert report[1] == ('-------------------------------------------------------------------')
+    assert report[2] == ('John Wick                  $    1000.00           1  $     1000.00\nJohn Smith'
+                         '                 $     670.00           3  $      223.33\nJane Doe             '
+                         '      $     570.00           3  $      190.00\nJason Bourne               $   '
+                         '  380.00           2  $      190.00\nGI Jane                    $     210.00   '
+                         '        2  $      105.00\n')
```
hosungs, line 22: Minor: Testing on printed strings can be very fragile, so it's rarely done. Output validation is not easy, and maybe it's outside the unit testing scope. But I do see some value in this anyway, and understand your intention.

students/shawn_degraw/session06/test_mailroom.py
```diff
@@ -0,0 +1,76 @@
...
+    testfile_johnwick = Path(path.join(letterdirectory, "John_Wick.txt"))
+    testfile_gijane = Path(path.join(letterdirectory, "GI_Jane.txt"))
+    testfile_johnsmith = Path(path.join(letterdirectory, "John_Smith.txt"))
+    testfile_jasonbourne = Path(path.join(letterdirectory, "Jason_Bourne.txt"))
+
+    assert path.isfile(testfile_janedoe) == True
+    assert path.isfile(testfile_johnwick) == True
+    assert path.isfile(testfile_gijane) == True
+    assert path.isfile(testfile_johnsmith) == True
+    assert path.isfile(testfile_jasonbourne) == True
```
hosungs, line 59: Great to see you are validating files are created! I wonder, though, if your test code delete all these files before `mailroompart4.sendletters()` is called (because what if the files were already there from the previous run, and this time it failed to generate the files).

One minor thing is that there's no need to put `== True` at the end of the `assert` statements. It's redundant.

hosungs (approved):
Looks good. Good job!

PR #79, opened 2018-11-14
Session06, mailroom testing

guoguoying2013:


students/YingGuo/session06/test_mailroom4.py
```diff
@@ -0,0 +1,57 @@
...
+
+"""session06, write test code for mailroom"""
+
+from mailroom4 import report
+from mailroom4 import all_letter
+from mailroom4 import donor_list
+from mailroom4 import letter_print
+from mailroom4 import add_new_donor
+from mailroom4 import letter_all
+from mailroom4 import all_letter
```
hosungs, line 11: Minor: You can import multiple names in one `from ... import ...` statement by enumerating all the imported names. E.g.,
```
from mailroom4 import report, all_letter, donor_list, letter_print, add_new_donor, letter_all, all_letter
```
Also, even though we learned that `from a_module import *` is discouraged, it's OK to use that if it's guaranteed not to cause any confusion, like in a test case. So I'd say it's OK to use `from mailroom4 import *` in this case.

students/YingGuo/session06/test_mailroom4.py
```diff
@@ -0,0 +1,57 @@
...
+    total_donation = 8888
+    expected = "Dear Ying, Thank you for your total donation of 8888"
+    actual = letter_all(name, total_donation)
+    assert expected == actual
+
+#test all_letter which creats file
+def test_created_file():
+    data = {"Tom":[100,200]}
+    all_letter(data)
+    assert os.path.isfile("Tom")
```
hosungs, line 57: You found it! One minor thing is that it's good to remove this generated file, because otherwise, this old file might affect the next test run (what if the next run fails to generate the file, and yet not caught because of this old file).

hosungs (approved):
Looks good overall. Good job!

PR #81, opened 2018-11-14
PR for assignment 6 

yushusongmsazure:
Modularize mailroom program and create unit tests to fully test the functions.

### DIFF
--- a/students/yushus/session05/mailroom3.py
+++ b/students/yushus/session05/mailroom3.py
@@ -8,6 +8,7 @@ Mailroom3 assignment
 
 import os
 import random
+import re
 import sys
 
 donor_db = {"William Gates, III": [653772.32, 12.17],
@@ -46,7 +47,6 @@ def send_a_thank_you():
 
     # maintain a donor list which contains donor names only
     donors = [d.lower() for d in donor_db.keys()]
-    name = ""
 
     while True:
         name = input("Who do you want to send this to, use a full name: ")
@@ -80,7 +80,9 @@ def send_a_thank_you():
 def send_letters_to_all():
     for name in donor_db:
         try:
-            with open(os.path.join(os.getcwd(), f"{name}.txt"), "w") as file:
+            # replace whitespace , * < > / \ ? % : | " . with underscore _
+            file_name = re.sub(r'[\s+|,|*|<|>|/|\\|?|%|:|\||"|.]', '_', name)
+            with open(os.path.join(os.getcwd(), f"{file_name}.txt"), "w") as file:
                 # Take the last donation made by the donor
                 file.write(format_letter().format(name=name, amount=donor_db[name][-1]))
         except FileNotFoundError as fnf_error:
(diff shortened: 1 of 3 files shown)


students/yushus/session05/mailroom3.py
```diff
@@ -80,23 +80,28 @@ def send_a_thank_you():
 def send_letters_to_all():
     for name in donor_db:
         try:
-            with open(os.path.join(os.getcwd(), f"{name}.txt"), "w") as file:
+            # replace whitespace , * < > / \ ? % : | " . with underscore _
+            file_name = re.sub(r'[\s+|,|*|<|>|/|\\|?|%|:|\||"|.]', '_', name)
```
hosungs, line 84: Cool regex usage!

students/yushus/session06/mailroom4.py
```diff
@@ -0,0 +1,140 @@
...
+        donor_db[name].append(amount)
+        return True
+    else:
+        raise ValueError('Donation has to be more than 0')
+
+# Add donor if not exist
+def update_donor(name, donor_db):
+
+    # Check if name is null or empty
+    if not name or name.isspace() or name is None:
```
hosungs, line 45: Minor: `name is None` is included in `not name`, so `name is None` is not necessary.

students/yushus/session06/test_mailroom4.py
```diff
@@ -0,0 +1,59 @@
...
+                                    ["Bill Gates", 300, 2, 150]]
+
+def test_update_donor_success():
+    donor_db = {"Bill Gates": [100, 200],"Jack Ma": [500, 500]}
+    update_donor("Yushu Song", donor_db)
+    assert "Yushu Song" in donor_db.keys()
+
+def test_update_donor_failure():
+    with pytest.raises(ValueError, match=r'Donor name.*'):
+        update_donor("", {})
```
hosungs, line 28: Good test that validates that your code raises an expected exception!

students/yushus/session06/test_mailroom4.py
```diff
@@ -0,0 +1,59 @@
...
+def test_get_file_name_new_dir_success():
+    try:
+        path = os.path.join(os.getcwd(), 'temp')
+        file_name = "Yushu*Song"
+        assert os.path.join(path, "Yushu_Song.txt") == get_file_name(path, file_name)
+    except Exception as ex:
+        print(ex)
+    finally:
+        # Clean up resource afterwards
+        shutil.rmtree(path)
```
hosungs, line 54: Great to see that you are cleaning up! However, I'd say it's a little unexpected that the function `get_file_name()` creates a directory if it's not there yet. It seems kind of an unexpected side effect.

It'd have been better if the directory was created in the actual file creation function. I know that you didn't want to pass the directory path to that function, but just pass the full path of the file only, which requires the directory to be created, so I suppose that's why you created the directory somewhere, but I'd say `get_file_name()` is not a good place to create the directory. Also, you could have run the code that generates all the letter files and validates that those files are really generated, and after that, remove the entire directory. Just a thought. No need to change your code at this point.

hosungs (approved):
Good job!

PR #82, opened 2018-11-16
Session 6 assignments by David Arasim

davidarasim:


students/dave_arasim/session06/test_mailroom4.py
```diff
@@ -0,0 +1,97 @@
...
+            textfile_found = True
+
+            textfile_content = textfile.read()
+
+        textfile.close()
+    except FileNotFoundError:
+        textfile_content = ''
+
+    assert textfile_found == True
+    assert textfile_content == form_text_out
```
hosungs, line 97: Good test to validate the file is generated with the expected content!

Just a minor comment, it's very fragile to compare formatted string outputs with expected strings (which you did in this test suite in many places), because even a single character difference (like an additional blank space) could cause the test failure. In fact, it's not easy to validate input/output, so mostly we end up separating processing logic with I/O and perform unit-testing on the processing logic only. Of course we know that this specific program needs lots of string processing, so it's kind of unavoidable, but it's still worth considering and trying to refactor code for better unit-testing (separate print() and file write into separate functions). No need to change, just FYI.

hosungs (approved):
Looks good overall. Thanks!

PR #83, opened 2018-11-17
late assignments 

ikemerrixs:
fibonacci series
Mailroom Part 4


students/merrix/session05/mailroom.py
```diff
@@ -0,0 +1,102 @@
...
+          "4 - To Exit",
+          ">>> "))
+
+
+def view_donor_db():
+    print (donor_db.keys())
+    choose = input("enter a name from the list or a new name to donate: ").title()
+    if choose not in donor_db.keys():
+        add_donor_db(choose)
+    elif choose in donor_db.keys():
```
hosungs, line 28: Minor: This `elif` condition is redundant, because it's just the negation of the `if` condition, and `else` alone is enough.

students/merrix/session05/mailroom.py
```diff
@@ -0,0 +1,102 @@
...
+        num_of_gift = len(donations)
+        total_given = int(sum(donations))
+        average_gift = int(total_given/num_of_gift)
+        print ('{0:<15}   {1:^10}     {2:^6}     {3:8}'.format(name, total_given, num_of_gift, average_gift))
+        
+
+def send_letters():
+    for donor_name in donor_db:
+        try:
+            letter = open('{}.txt'.format(donor_name), "w+")
```
hosungs, line 68: Minor: Use of `with` is recommended: `with open(...) as letter:`

hosungs (commented):
Thanks for your work and PR. Left a few comments, but they are not important.

What's important is to meet the assignment requirements. For mailroom part 3, you have one exception handling code, so I'd say that requirement is met. However, I don't see any list comprehension (which is the other requirement), so I think you should add some. I guess it's not straightforward what to change to a comprehension, and you might need to decompose some of your functions so that you can create a list using comprehension and use the created list for later processing (e.g., report formatting).

I also see that your `session06/mailroom4.py` is the same as `session05/mailroom3.py`, so I assume it's work in progress. Please add unit tests to your `session06` directory. Note that that may require you to restructure and reorganize your code into more functions so that you can add unit tests at least for certain functions.

I'll leave this PR open as such. Keep pushing your new commits. Thanks!

hosungs (commented):
Checked the PR after the latest commits on Nov. 26, but only the `strformat_lab.py` was added, nothing else. Mailroom assignments are more important than other labs, please prioritize on mailroom assignments. Make sure to submit the `strformat_lab.py` to the corresponding Canvas assignment dropbox so that our TA can grade.

Will keep this PR open for more work on mailroom assignments.

hosungs (commented):
Checked the PR again after a bunch of email regarding new commits, but I don't see any new code that actually tackles any assigned problems. And it seems like you are trying multiple assignments at once, without finishing the first one first and moving on to the next, which makes PR closure quite difficult. I'll have to keep this PR open as such, and just make sure you put proper comments in your Canvas submissions (I assume you are making those--otherwise, please do).

students/merrix/session06/mailroom4.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+
+def user_Report(donor_db):
+    print('Donor Name  | ',  'Total Given | ', 'Num Gift  | ', 'Average Gift')
+    for name, donations in donor_db.items():
+        num_of_gift = len(donations)
+        total_given = int(sum(donations))
+        average_gift = int(total_given/num_of_gift)
+        return('{0:<15}   {1:^10}     {2:^6}     {3:8}'.format(name, total_given, num_of_gift, average_gift))
```
hosungs: Placing `return` inside the `for` loop will cause the loop to terminate immediately after the first iteration, and that'll not create the entire report for all donors. Is this what you intended? If you intended this function to create a report for only one donor, why create a loop in the first place? You need to think through your intention and logic here and clarify/fix this code.

hosungs (changes requested):
There's some test code, so I'd say the mailroom_pt4's requirements are met at least in some degree, but there's some questionable logic in the `user_Report()`. Need to fix that correctly in order for us to accept this as an acceptable submission.

hosungs:
Closing this PR, as the class is over for the quarter. Any further request for feedback can be made using other PRs, but the ongoing commits that are piling up on this PR are not helpful at all, so closing.

PR #84, opened 2018-11-18
Added Lesson06

nallarun10:


students/arun_nalla/session06/test_Mailroom_Part4.py
```diff
@@ -0,0 +1,49 @@
+#!usr/bin/env python
+""" Testing Mailroom functions using pytest"""
+
+from Mailroom import get_report
+from Mailroom import donors_dict
+from Mailroom import donors_list
+from Mailroom import add_donors
+from Mailroom import email_message
+from Mailroom import path_to_store
+from Mailroom import add_donation
```
hosungs, line 10: Minor: You can combine all these imports into one line: `from Mailroom import get_report, donors_dict, donors_list, ...`. Also it's not that bad to use `from Mailroom import *` in a case like this (unit testing, that's very confined), even though we learned that `from ... import *` is not a good practice.

students/arun_nalla/session06/test_Mailroom_Part4.py
```diff
@@ -0,0 +1,49 @@
...
+    assert add_donation(name, demo_dict, donation) == expected_dict
+    assert 1 in demo_dict['Arun']
+    assert len(demo_dict['Arun']) == 1
+    assert len(demo_dict['Abc']) == 2
+    assert 1 not in demo_dict['Xyz']
+    assert 10 not in demo_dict['Abc']
+
+def test_get_report():
+    expected_sorting = [('Xyz', 10, 1, 10.0), ('Abc', 3, 2, 1.5), ('Arun', 1, 1, 1.0)]
+    assert get_report(demo_dict) == expected_sorting
```
hosungs, line 42: Minor: The success of this test case depends on the execution of `test_add_donations()` (otherwise it'll fail), and I'm not sure how pytest guarantees that. In fact, I think it's safe not to assume such a specific ordering.

Probably the right approach is to initialize the `demo_dict` whenever a test case is executed and let the test case work on the re-initialized `demo_dict`. In fact, there are pytest constructs that allow such initialization (like xunit-style setup/teardown construct, or pytest fixtures), but we didn't cover them in class. At least we can have a separate function that initializes needed variables that can be called by every test case at the beginning. Just FYI for the future and no need to change your code.

students/arun_nalla/session06/test_Mailroom_Part4.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_path_to_store():
+    name = random.choice(list(demo_dict))
+    assert path_to_store(name) == os.path.join(os.getcwd(), 'Mailroom', 'email_{}.txt'.format(name))
+    expected_message = "Dear {},\nThank you for generous donation of ${:.2f}. It will be put to a good cause.\nSincerely,\nThe Team".format(name,sum(demo_dict[name]))
```
hosungs, line 48: Another minor point: It'd be good to define this formatting string as a variable so that it can be reused. Otherwise, this test will fail if your formatting string changes in your `mailroom.py`, but not here (and maintaining the same duplicates is tedious).

hosungs (approved):
Left a few comments, no need to change your code for full credits, but feel free to update your code as suggested. Will merge this PR for the future assignment. Feel free to include any commits on this assignment in your future PR, if you'd like. Thanks!

PR #85, opened 2018-11-21
session05: mailroom03, revised for feedback

jjakubiak:
I revised mailroom03 for the feedback I received regarding adding list comprehension. I like to get a review of the revisions before I submit mailroom04. Thanks!

### DIFF
--- a/students/jjakub/session05/mailroom03.py
+++ b/students/jjakub/session05/mailroom03.py
@@ -1,5 +1,12 @@
+#!/usr/bin/env python3
 import sys
 import tempfile
+
+"""
+Class: Python 210 B, Au2018
+Exercise: Session 05, Mailroom Part 03
+Student: Jason Jakubiak
+"""
 
 user_options = {
     "Send": "Send a Thank You to a single donor",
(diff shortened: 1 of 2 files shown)


students/jjakub/session05/mailroom03.py
```diff
@@ -32,43 +46,54 @@ def send_thanks():
...
         except ValueError:
-            print("\n Warning: Please enter an integer!\n")
+            print("\n Warning: Please enter an integer! \n")
+
+
+def donor_calc(donor_db):
+    donor_sum = [
+            (donor.capitalize(), sum(amt), len(amt), sum(amt)/len(amt)) 
+            for donor, amt in donor_db.items()
+            ]
```
hosungs, line 60: Definitely a good list comprehension, though unfortunately `sum(amt)` is computed twice (unnecessarily repeated)... Let's focus more on the pedagogical aspect and call this fine, though in production, you wouldn't want to repeat the same computation twice unnecessarily. I'm not sure if this can be automatically optimized by Python runtime, and my guess is that it is not...

hosungs (approved):
Looks good. Thanks for following up on feedback!

PR #86, opened 2018-11-21
Beta release of Mailroom4 and Unit Test

bmnguyen:
Session 6 mailroom 4 final


students/arun_nalla/session06/Mailroom.py
```diff
@@ -0,0 +1,103 @@
+#!usr/bin/env python
```
hosungs: Umm... This file is in someone else's directory, and not sure how this happened. Strangely there's no merge conflict, but this can't be merged, because it'll overwrite Arun's files. We'll need to fix this somehow...

students/arun_nalla/session06/test_Mailroom_Part4.py
```diff
@@ -0,0 +1,49 @@
+#!usr/bin/env python
```
hosungs: Same here (this file is in Arun's directory)

students/brandon_nguyen/session06/mailroom4.py
```diff
@@ -65,21 +69,77 @@ def create_rpt():
...
     """
-    This function returned a sorted list of by on order amount3
+    This function returned a sorted list of by on order amount
     """
     dblist = []
     dbSum = []
     # quick way to convert back to list to reuse existing code.
     # list comprehension
-    [dblist.append((i, j)) for i, j in donor_db.items()]
+    [dblist.append((i, j)) for i, j in dict_db.items()]
```
hosungs, line 82: Still not very desirable list comprehension, though it'll work. Desired style is:
```
dblist = [(i, j) for i, j in dict_db.items()]
```
In fact, in this case, probably `dblist = list(db.items())` would be enough.

Same for the next line.

students/brandon_nguyen/session06/mailroom4.py
```diff
@@ -65,21 +69,77 @@ def create_rpt():
...
+    input_val is a list with two set of value, donor_name, donation_value.
+    db_dict is a replication of donor_db
+    """
+    # existing people add to donation
+    if input_val[0] in dict_db:
+        new_donation = dict_db.get(input_val[0])
+        new_donation.append(input_val[1])
+        dict_db[input_val[0]] = new_donation
+    else:
+        dict_db.update({input_val[0]: [input_val[1]]})
```
hosungs, line 101: Not related to unit testing, but still related to dictionary coding style. Lines 96-101 could have been simplified to `dict_db[input_val[0]].append(input_val[1])` if `dict_db` is defined as `defaultdict(list)`.

students/brandon_nguyen/session06/test_mailroom4.py
```diff
@@ -0,0 +1,62 @@
...
+
+input_db = {'FirstNameA LastNameA': [1, 1, 1, 11],
+            'FirstNameB LastNameB': [2, 2, 2, 22],
+            'FirstNameC LastNameC': [3, 3, 3, 33],
+            }
+
+TEST_LETTER = "\n".join(("", "Dear {Name},", "", "Thank you for your "
+                         "kind donation of {LastAmnt:.2f}.",
+                         "It will be put to very good use.", "",
+                         "Sincerely, ", "", "-The Team\n"))
```
hosungs, line 24: Isn't this the same as `E_FORMAT` defined in your `mailroom4.py`? If so, it's best to import and reuse it here (`from mailroom4 import E_FORMAT`), rather than duplicating it here.

students/brandon_nguyen/session06/test_mailroom4.py
```diff
@@ -0,0 +1,62 @@
...
+import sys
+import time
+import unittest
+import os
+from decimal import Decimal
+from datetime import datetime, date
+from mailroom4 import sort_sum4_report
+from mailroom4 import update_donation
+from mailroom4 import create_letters_text
+from mailroom4 import create_files
```
hosungs, line 13: Minor: Better combine lines 10-13 to `from mailroom4 import sort_sum4_report, update_dontaion, create_letters_text, create_files`. It's also OK to do `from mailroom4 import *` in a test suite that's dedicated to the `mailroom4.py` module.

students/dave_arasim/session06/cigar_party.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
```
hosungs: Someone else's file. Same for all the remaining files. Not sure how this happened, but we'll need to remove all these files (by others) in your PR before merging it. I'll think about how to do that offline...

hosungs (commented):
Left some comments. Can't merge because there are other students' files added in this PR (not sure how this happened). I'll think about how to fix that and let you know, or we'll need to do that together in person.

hosungs (commented):
I see test cases in your `session06/test_mailroom4.py`, so I think this is no longer the beta release. Therefore, I'll merge/close this PR, so that your future commits don't pile up here. Thanks!

PR #87, opened 2018-11-26
Session07 html render

motians:
Tests seem to pass and run_html_render seems to print rendered html correctly for steps 8 & 9.

Looking for any comments on pythonic coding and if my results cover the spirit of the assignment.

### DIFF
--- a/students/shawn_degraw/session07/html_render.py
+++ b/students/shawn_degraw/session07/html_render.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+    tag_name = "html"
+
+    def __init__(self, content=None):
+        if content:
+            self.html_content = [content]
+        else:
+            self.html_content = []
+
+    def append(self, new_content):
+        self.html_content.append(new_content)
+
+    def render(self, out_file):
+        out_file.write("<{}>\n".format(self.tag_name))
+        for htmlitem in self.html_content:
+            if type(htmlitem) == str:
+                out_file.write("{}\n".format(htmlitem))
+            else:
+                htmlitem.render(out_file)
+        out_file.write("</{}>\n".format(self.tag_name))
+
+
+class Html(Element):
+    tag_name = "html"
+
+
+class Body(Element):
+    tag_name = "body"
+
+
+class P(Element):
+    tag_name = "p"
+
+
+class Head(Element):
+    tag_name = "head"
+
+
+class OneLineTag(Element):
+    tag_name = "title"
+
+    def render(self, out_file):
+        out_file.write("<{}>".format(self.tag_name))
+        for htmlitem in self.html_content:
+            out_file.write("{}".format(htmlitem))
+        out_file.write("</{}>\n".format(self.tag_name))
+
+
+class Title(OneLineTag):
+    tag_name = "title"
--- a/students/shawn_degraw/session07/html_render.py
+++ b/students/shawn_degraw/session07/html_render.py
@@ -5,9 +5,11 @@ A class-based system for rendering html.
 """
 
 
-# This is the framework for the base class
 class Element(object):
+    """ Default Element class for rendering html """
     tag_name = "html"
+    indent = 3
+    cur_ind = 0
 
     def __init__(self, content=None, **kwargs):
         self.html_content = ([content] if content else [])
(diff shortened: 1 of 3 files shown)


students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+        self.html_content.append(new_content)
+
+    def render(self, out_file, cur_ind=""):
+        if self.header_arg:
+            self.writetag(out_file, cur_ind + "<{}")
+            for k, v in self.header_arg.items():
+                out_file.write(" {}=\"{}\"".format(k, v))
+            out_file.write(">\n")
+        else:
+            self.writetag(out_file, cur_ind + "<{}>\n")
```
hosungs: Looks like the `if` condition and this `else` block are redundant, because if `self.header_arg` is empty, then the `for` loop will not iterate at all, thus giving the same output for the `else` case, even with the code in the `if` block.

students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+    """ subclass of Element for rendering tags with no content """
+    tag_name = ""
+
+    def __init__(self, content=None, **kwargs):
+        if content:
+            raise TypeError('Error: Content not allowed for tag.')
+        else:
+            self.html_content = []
+
+        self.header_arg = (kwargs if kwargs else {})
```
hosungs: You are not reusing the superclass initializer, but redefining the `self.html_content` and `self.header_arg` here separately. It's better to use `super().__init__(content=content, kwargs=kwargs)`. You can see this instruction in the tutorial. Maybe I didn't emphasize how to invoke the superclass initializer from a subclass...

students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+        if self.header_arg:
+            self.writetag(out_file, cur_ind + "<{}")
+            for k, v in self.header_arg.items():
+                out_file.write(" {}=\"{}\"".format(k, v))
+            out_file.write(" />\n")
+        else:
+            self.writetag(out_file, cur_ind + "<{} />\n")
+
+
+class A(Element):
```
hosungs: `A` better be a subclass of `OneLineTag`, not of `Element`. See tutorial for the step.

students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+    """ subclass of Element for rendering a tags """
+    tag_name = "a"
+
+    def __init__(self, link=None, content=None):
+        if not link or not content:
+            raise TypeError('Error: Value cannot be null.')
+        else:
+            self.href_link = link
+            self.html_content = content
```
hosungs: You are not reusing the superclass capabilities, but re-implementing needed stuff separately here. Try to use `super().__init__(...)` properly with the passed params here (`link` and `content`). Tutorial shows you how.

students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind + "<a href=\"{}\">{}</a>\n".format(self.href_link, self.html_content))
```
hosungs: Again, it's best to reuse the `render()` method defined in the superclass (`OneLineTag`). If you had used the `super().__init__(...)` properly in this class's `__init__()`, then you wouldn't even have had to define this method at all, because the superclass's `render()` method just performs this as expected.

students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class H(OneLineTag):
+    """ subclass of OneLineTag for rendering h tags """
+    def __init__(self, level, content):
+        self.tag_name = "h{}".format(level)
+        Element.__init__(self, content)
```
hosungs, line 86: Yes, you did use the superclass's initializer here! You can do the same on the `A` class.

students/shawn_degraw/session07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class Html(Element):
+    """ subclass of Element for rendering html tags """
+    tag_name = "html"
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind + "<!DOCTYPE html>\n")
+        Element.render(self, out_file, cur_ind)
```
hosungs, line 95: Yes, this is the right way to reuse the superclass's `render()` method. You had to override `render()` here in the subclass, because of the additional line (`<!DOCTYPE html>`), but otherwise, everything else can/should be rendered through the superclass's `render()` method. Your `A` class will not even need to override `render()` in there, because there's nothing else that needs to be done (unlike in the case of `Html`).

hosungs (commented):
Definitely on the right track. A few changes are recommended. Other than those few, looks good!

I'll keep this PR open until the due is over. Thanks!

students/shawn_degraw/session07/html_render.py
motians: Step 6 indicates the 'A' should be a subclass of Element:  "You should be able to subclass from Element, and only override the __init__ — calling the Element __init__ from the A __init__"

students/shawn_degraw/session07/html_render.py
motians: Interesting.  I assumed the for loop would do something even when the header_arg was empty.  Took out the if/else and everything still worked fine!

students/shawn_degraw/session07/html_render.py
motians: I went ahead and used the OneLineTag and passed the link to the init and it seems to work correctly now without overriding the render.  I have been avoiding the tutorial and trying to do it myself.

class A(OneLineTag):
    """ subclass of OneLineTag for rendering a tags """
    tag_name = "a"

    def __init__(self, link=None, content=None):
        super().__init__(content, href="{}".format(link))

students/shawn_degraw/session07/html_render.py
motians: I think I have this corrected in the newest commit.

students/shawn_degraw/session07/html_render.py
hosungs: Hmm, I must have missed that statement in step 6, but that looks like a typo... I think what you did is indeed what's desired, so let's keep it that way (should be no deduction for this change). Thanks!

hosungs (approved):
Looks all good now. Will merge/close this PR as such. Thanks!

PR #88, opened 2018-11-27
PR for session07 HTML render

yushusongmsazure:
Finished 9 steps.


students/yushus/session07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+        return f"<{self.tag}{self._get_attributes()}>"
+
+    def _close_tag(self):
+        return f"</{self.tag}>"
+
+    def _get_attributes(self):
+        page_attributes = ""
+        if self.attributes:
+            for attribute in self.attributes:
+                page_attributes = f"{page_attributes} {attribute}=\"{self.attributes[attribute]}\""
```
hosungs: Minor: You are performing string concatenation using something similar to `s = s + append_stuff`, which works, but not recommended. The desirable approach is to create a list of component strings and then use string 'join()' method. In fact, it can all be done in one line using list comprehension:
```
return ' '.join(f'{k}={v}' for k, v in self.attributes.items())
```

students/yushus/session07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+class Html(Element):
+    tag = "html"
+    def render(self, out_file, cur_ind=""):
+        out_file.write(f"{cur_ind}<!DOCTYPE {self.tag}>")
+        out_file.write("\n")
+        super().render(out_file, cur_ind)
```
hosungs, line 58: Good job, reusing the superclass method!

students/yushus/session07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+    tag = "br"
+
+class Meta(SelfClosingTag):
+    tag  = "meta"
+
+class A(OneLineTag):
+    tag = "a"
+    def __init__(self, link, content=None, **attributes):
+        attributes["href"] = link
+        super().__init__(content, **attributes)
```
hosungs, line 108: Good job, reusing the superclass as much as possible! This way, you didn't need to redefine `A.render()` method at all (because the superclass's `render()` method does all the job for this already).

hosungs (commented):
Looks good! Left a few comments for your reference, and will leave this PR open until the due is over. Thanks!

students/yushus/session07/html_render.py
yushusongmsazure: Nice! Thanks Hosung!

hosungs (approved):
Additional changes look good. Merging/closing this PR for your future commits. Thanks!

PR #89, opened 2018-11-27
Session06 submission

jjakubiak:
Submission for mailroom04 and mailroom04 unit test

### DIFF
--- a/students/jjakub/session06/test_mailroom04.py
+++ b/students/jjakub/session06/test_mailroom04.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+"""
+Class: Python 210 B, Au2018
+Exercise: Session 06, Mailroom Part 04 Unit Test
+Student: Jason Jakubiak
+"""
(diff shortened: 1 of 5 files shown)


students/jjakub/session06/test_mailroom04.py
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+test_list_out = ["Name"]
+
+test_txt = (
+    "\nDear Name,"
+    "\nThank you for the donation of $9999.00."
+    "\nSincerely,"
+    "\nThe Mailroom Foundation\n"
+    )
```
hosungs, line 29: Minor: Not sure why you had this as a tuple (same in `mailroom04.py` above), but it's best to avoid unnecessary code duplication like this. Consider using one template string, and derive this compared string by using the template with the expected values.

hosungs (approved):
Looks good overall for week 6 assignment (mailroom part 4). Merging/closing this PR as such for future assignments.

PR #90, opened 2018-11-28
Session 07 PR

funnyletter:


students/rachel_s/session07/html_render.py
```diff
@@ -0,0 +1,157 @@
...
+            self.args = kwargs
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def _indenter(self, cur_ind):
+        output = ''
+        for i in range(cur_ind):
+            output += self.indent
+        return output
```
hosungs, line 32: Minor: Maybe simpler to say `return self.indent * cur_ind`?

students/rachel_s/session07/html_render.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def render(self, out_file, cur_ind=0):
+        out_file.write(self._indenter(cur_ind))
+        out_file.write(self._open_tag())
+        out_file.write('\n')
+        for item in self.contents:
+            try:
+                item.render(out_file, cur_ind=cur_ind+1)
```
hosungs, line 40: Good idea for indentation!

students/rachel_s/session07/html_render.py
```diff
@@ -0,0 +1,157 @@
...
+            out_file.write("\n")
+        out_file.write(self._indenter(cur_ind))
+        out_file.write(self._close_tag())
+
+    def _unpack_kwargs(self):
+        output = ''
+        for key, value in self.args.items():
+            output += ' {par}=\"{val}\"'.format(par=key, 
+            val=str(value))
+        return output
```
hosungs, line 53: Minor: Consider using a list comprehension for things like this, like `return ' '.join('{par}="{val}"'.format(par=key, val=str(value)) for key, value in self.args.items())`. Of course you'll still need to handle the leading blank space in this case, when self.args is not empty.

hosungs (approved):
Looks good for html renderer. Not sure why previous assignments are in this PR, and I didn't look at them anyway. Merging/closing this PR for future assignments. Thanks!

PR #91, opened 2018-11-29
Trigrams and mailroom pt 4 with unit tests (pytest)

ghost:


students/timm67/session04/trigrams.py
```diff
@@ -1,22 +1,179 @@
...
+                return None
+            while(line):
+                line = line.replace('--', ' ')
+                line = filter_trigram(line, exceptions)
+                if(len(line) == 0):
+                    try:
+                        line = fd.readline()
+                    except IOError:
+                        print("I/O Error with [{0}] readline".format(in_filename))
+                    continue
```
hosungs, line 83: Minor: This `if` block seems unnecessary...? If `len(line) == 0`, then it's EOF (end-of-file), and no more `.readline()` will be possible (maybe I'm wrong). No matter what, it'll just print the error message and continue anyway, so doesn't seem necessary.

students/timm67/session04/trigrams.py
```diff
@@ -1,22 +1,179 @@
...
+        key = (word_list[i], word_list[i+1])
+        if key in trigrams_dict.keys():
+            trigrams_dict[key].append(word_list[i+2])
+        else:
+            trigrams_dict[key] = [word_list[i+2]]
+
+
+def build_trigrams_list(words_list, trigrams_dict):
+    for line_list in words_list:
+        build_trigrams(line_list, trigrams_dict)
```
hosungs, line 28: Minor: This function doesn't seem to be called from anywhere, so it better be removed.

students/timm67/session05/mailroom_part4.py
```diff
@@ -0,0 +1,192 @@
...
+        donor_amt = input("Enter donation value: ")
+        try:
+            donor_amt = float(donor_amt)
+            break
+        except ValueError:
+            print("Please enter a numeric donation value")
+
+    # Not optimal for optimization via list comprehension
+    for donor in donors.keys():
+        if donor[0] == donor_fname and donor[1] == donor_lname:
```
hosungs, line 44: Your `donors` is a dict, but you are doing a sequential search (like on a list), which is not desired. I see that your keys are tuples, so you could have done something like the following:
```
if (donor_fname, donor_lname) in donors:
    donors[(donor_fname, donor_lname)].append(donor_amt)
else:
    donors[(donor_fname, donor_lname)] = [donor_amt]
```
Better than this is to use `defaultdict(list)` for `donors`, and just avoid the `else` part all together (no need to check `if` condition if `donors` is a `defaultdict`).

students/timm67/session05/mailroom_part4.py
```diff
@@ -0,0 +1,192 @@
...
+    print("[1] Send a Thank you to a single donor")
+    print("[2] Create a report")
+    print("[3] Send letters to all donors")
+    print("[9] Exit Mailroom")
+    pass
+
+
+def exit_mail():
+    global exitMail
+    exitMail = True
```
hosungs, line 161: Minor: Not very desirable to rely on a global variable. Could return some value and let the main loop receive/detect the special value for exit. Also could just `exit()` here.

students/timm67/session05/mailroom_test.py
```diff
@@ -0,0 +1,64 @@
...
+     ('Rex', 'Miller'): [20.00, 20.00, 5.00],
+     ('Tony', 'Blake'): [20.00, 20.00, 10.00]
+     }
+
+thankyou_file_list = []
+
+
+def test_report():
+    outlines = create_report_task()
+    assert(5 == len(outlines))
```
hosungs, line 31: No need to change, but it would have been better to check the content of the `outlines` list as well, not just the number of items.

students/timm67/session05/mailroom_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_thankyou_files():
+    global thankyou_file_list
+    thankyou_file_list = send_thankyou_multiple_donors_task()
+    for test_file in thankyou_file_list:
+        try:
+            test_file_path = Path(test_file)
+            assert(test_file_path.is_file())
```
hosungs, line 40: No need to change, but it'd be better to check the content of the generated files, not just checking the file. Also, I'm not sure if you remove all these files after each test is run, because previously generated files might be visible here, when potential new code change might break the functionality.

hosungs (approved):
Glad to see your progress. I think your `session05/mailroom_part4.py` can be considered for both session 5 (which is mailroom part 3) and session 6 (which is part 4). Please do make sure to submit the files on both Canvas assignments so that TA can grade them both.

I'll merge/close this PR, so that your new code for session 7 & 8 will not pile up here. Thanks.

PR #92, opened 2018-11-29
lession07, 08 homework and lession 09 Ying

guoguoying2013:


students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+
+# This is the framework for the base class
+class Element():
+    tag_name = "html"
+    indent = "    "
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        if content is not None:
```
hosungs: Minor: This line should be just `else`.

Better yet, the whole `if`-`else` block better be simply `self.content = [] if content is None else [content]`.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+            self.content = [content]
+        self.kwargs = kwargs 
+
+
+    def append(self, new_content):
+        self.content.append(new_content)
+
+    def write_html(self):
+        with open("output_html","w") as f:
+             f.write(self.content)
```
hosungs: Minor: Do you ever use this method (`write_html`) anywhere? I don't think this method was required in the instructions or tutorial...? No big deal, and I might be mistaken again... Just wondering...

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+             
+    #what does ind mean in the test_html? why it is needed?
+    def render(self, out_file, cur_ind=""):
+        if cur_ind:
+            out_file.write(cur_ind)
+        else:
+            pass
```
hosungs: Minor: No need to put `else: pass`. It's of no use, and that's exactly what happens when you don't enter that code.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+
+        for content in self.content:
+            out_file.write("<{}>\n".format(self.tag_name))
+
+            if cur_ind:
+                out_file.write(cur_ind)
+            else:
+                pass
```
hosungs: Ditto.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+            #
+            out_file.write(self.indent)
+            try:
+                content.render(out_file)
```
hosungs: Not sure if you are getting indented output, even though you are not passing any `cur_ind` here? Does your indentation work well?

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+                out_file.write(content)
+            out_file.write("\n")
+            
+            out_file.write("</{}>\n".format(self.tag_name))
+
+
+#a subclass of Element with tag body
+class Body(Element):
+    tag_name = "body"
+    indent = "    "*2
```
hosungs: This is not right... You shouldn't hard-code the indentation level. It should have been programmatically handled.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+
+#subclass for paragraph
+class P(Element):
+    tag_name = "p"
+    indent = "    "*3
```
hosungs: Ditto. You can't assume the `p` element appears only at level 3. It can be placed in any deeper levels.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+        for content in self.content:
+            if self.kwargs is not None:
+                open_tag = ["<{}".format(self.tag_name)]
+                for x in self.kwargs:
+                    open_tag.append(x)
+                    open_tag.append("=")
+                    open_tag.append('"{}"'.format(self.kwargs[x]))
+                    open_tag.append(" ")
+                open_tag.append(">\n")
+                out_file.write("".join(open_tag))
```
hosungs, line 35: This attribute-rendering logic is needed everywhere, not just in the `p` element only, so this code should be moved up in the `Element` class.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,162 @@
...
+        open_tag = ["<{} ".format(self.tag_name)]
+        if self.kwargs:
+            for x in self.kwargs:
+                open_tag.append(x)
+                open_tag.append("=")
+                open_tag.append('"{}"'.format(self.kwargs[x]))
+                out_file.write("".join(open_tag))
+                out_file.write(" ")
+        else:
+            out_file.write("".join(open_tag))
```
hosungs, line 99: You had to duplicate the same attribute-rendering code here as well, because you didn't put the earlier code in the `Element` class. The key point is to avoid this kind of code duplication by utilizing inheritance and reusing the inherited methods as is.

hosungs (commented):
Left some comments, and it'd be great if you can address them. If not, we can just accept this submission as is and TA will grade as such.

No matter how, this PR can't be merged because there are a lot of merge conflicts, with lots of files included in this PR, that are from outside your student directory (`students\YingGuo`), so I can't merge/close this PR anyway. We'll need to fix this in class or at the office hours. I can craft some recipe you can try to fix this issue, but that might be more confusing, so let's plan on fixing this together. Thanks.

students/YingGuo/lession07/html_render.py
guoguoying2013: Nope it didn't work as I want. I am going to work on it and re-submit by Tuesday night. Hope I can figure it out.

students/YingGuo/lession07/html_render.py
hosungs: It's good that you moved your attribute-rendering logic up here, but I see a minor issue, where the first attribute might not be separate from the tag name (there's no blank space).

Another minor thing is that the 4 `open_tag.append(...)` calls in the `for` loop better be consolidated into just 1: `open_tag.append(' {}="{}"'.format(x, self.kwargs[x]))`

students/YingGuo/lession07/html_render.py
hosungs: There's still some code duplication here (the attribute-rendering logic). I know not the whole code can be reused as is (because of that slight difference at the closing tag), but you can reuse most of code by breaking things up into the opening tag part (including all the attributes) and the closing tag part.

students/YingGuo/lession07/html_render.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, content = None, **kwargs):
+        if content is not None:
+            raise TypeError("SelfClosingTag can not contain any content")
+        super().__init__(content=content, **kwargs)
+
+class A(SelfClosingTag):
+    tag_name = "a"
+
+    def __init__(self,link,content, **kwargs):
+        super().__init__(content, **kwargs)
```
hosungs, line 122: You are losing the `href={link}` attribute here!

students/YingGuo/session08/circle.py
```diff
@@ -0,0 +1,64 @@
...
+    @property
+    def diameter(self):
+        self.D = self.radius*2
+        return self.D
+
+    @diameter.setter
+    def diameter(self, d):
+        self.radius = d/2
+
+    #The user should not be able to set the area. How to test this?
```
hosungs, line 22: You can try something like:
```
c = Circle(3)
c.area = 5
```
And wrap the 2nd line with the `with pytest.raises(...):` with the observed exception type on the failed execution.

students/YingGuo/session08/circle.py
```diff
@@ -0,0 +1,64 @@
...
+    @property
+    def area(self):
+        self.a = pi*(self.radius**2)
+        return self.a
+    
+    @classmethod
+    def re_construct(cls, D):
+        cls.radius = D/2
+        return cls.radius
```
hosungs, line 31: Not sure if this is the right thing to do. I believe the `@classmethod` here is to create an alternative constructor, so it should return an instance of `cls`, not its attribute. You should just have `return cls(D/2)`.

hosungs (commented):
Still seeing some minor issues in both html_render and circle, so I'll leave this PR open just in case you keep fixing those issues.

students/YingGuo/session09/cli_main.py
```diff
@@ -0,0 +1,36 @@
...
+            1. add new donor\n\
+            2. add donation to existing donor\n\
+            3. full report\n\
+            4. send thank you\n\
+            5. quit"))
+
+            if user_input == 1:
+                input_name = input("What's the name of new donor?")
+                dc.add_new_donor(input_name)
+                print(dc.__init__())
```
hosungs: Hmm... Why are you doing this? `dc.__init__()` doesn't return anything, so nothing (except `None`) will be printed. Besides, calling `dc.__init__()` will reset the underlying `dc.__donors` dictionary that stores the added donors. Is that really what you wanted? You should have tested more involved workflow, not just individual features separately.

hosungs (commented):
Left a comment, after oo_mailroom commits were pushed. It also looks like you haven't yet implemented the thank you letter file saving feature. Please address those.

hosungs (approved):
Merging the PR, after seeing you completed the file writing feature. Thanks.

PR #93, opened 2018-12-02
Added Session07 - html_render

nallarun10:


students/arun_nalla/session07/html_render.py
```diff
@@ -0,0 +1,137 @@
...
+
+class P(Element):
+    tag = 'p'
+
+class Head(Element):
+    tag = 'head'
+
+class OneLineTag(Element):
+    tag = 'title'
+
```
hosungs, line 77: Minor: You may want to override `__init__()` here, so that `None` for `content` or any non-string `content` is not accepted for `OneLineTag` instances.

students/arun_nalla/session07/html_render.py
```diff
@@ -0,0 +1,137 @@
...
+class SelfClosingTag(Element):
+
+    def __init__ (self, content = None, **kwargs):
+        if content is not None:
+            raise TypeError ('Cannot have content')
+        super().__init__(content=content, **kwargs)
+
+    def _open_tag(self, cur_ind = ''):
+        open_tag = '<{}{}/>'.format(self.tag, self.attributes())
```
hosungs, line 98: Minor: I think there should be a blank space before `/>` all the time...?

students/arun_nalla/session07/html_render.py
```diff
@@ -0,0 +1,137 @@
...
+        return cur_ind + open_tag
+
+    def render(self, outfile, cur_ind = ''):
+        open_tag = '<{}{}/>'.format(self.tag, self.attributes())
+        outfile.write(cur_ind)
+        outfile.write(open_tag)
```
hosungs, line 104: Why duplicate code (same as in `_open_tag()`)? Looks like all this could be easily done by `outfile.write(self._open_tag(cur_ind))`...?

hosungs (approved):
Looks good overall, except a few minor comments I mentioned in this PR. Feel free to ignore them or push a new commit addressing them. Closing this PR as is, because the due was over already. Thanks!

PR #94, opened 2018-12-02
Circle and Sphere classes with tests for session08

motians:


students/shawn_degraw/session08/circle.py
```diff
@@ -0,0 +1,81 @@
...
+        return round(pi * self.radius ** 2, 6)
+
+    def __str__(self):
+        return "Circle with radius: {}".format(self.radius)
+
+    def __repr__(self):
+        return "Circle({})".format(self.radius)
+
+    def __add__(self, other):
+        return type(self)(self.radius + other.radius)
```
hosungs, line 49: Interesting to see this. I thought `Circle(self.radius + other.radius)` would be just fine and desirable, but it looks like this makes more sense or desirable (unless there's some Python gotcha I don't know about)... Thanks for being inquisitive!

students/shawn_degraw/session08/test_circle.py
```diff
@@ -0,0 +1,262 @@
...
+def test_setarea():
+
+    c = Circle(2)
+
+    try:
+        c.area = 12
+    except AttributeError:
+        assert(True)
+    else:
+        assert(False)
```
hosungs: I have to admit that I don't remember the syntax and I looked this up, but a better pytest syntax for this would be as follows:
```
    with pytest.raises(AttributeError):
        c.area = 12
```
Our course material (and associated example code) definitely talks about this syntax, but we rarely used this, so it might feel remote. Hopefully this time all of us remember this syntax... :)

students/shawn_degraw/session08/test_circle.py
```diff
@@ -0,0 +1,262 @@
...
+    assert c.diameter == 8
+
+
+# Step 6 tests
+
+def test_repr():
+
+    c = Circle(5)
+    d = repr(c)
+    assert d == 'Circle(5)'
```
hosungs: Not bad, but probably `assert eval(d) == c` might have been more desirable.

hosungs (approved):
Looks good. Just left a few minor comments. I'd not hesitate merging/closing this PR, but I'll leave this open until the due is over. Thanks for your diligence and good work!

PR #95, opened 2018-12-04
circle/sphere and more html_render work

ghost:


students/timm67/session08/circle.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+
+from math import pi
+
+
+class Circle():
+    _radius = float(0.0)
+    _diameter = float(0.0)
+    _area = float(0.0)
+    _circumference = float(0.0)
```
hosungs: Not sure why you defined these class attributes, when they better/should be instance attributes that should be defined in `__init__()` (which you did). Probably these lines (13-16) can/should be removed.

students/timm67/session08/circle.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def diameter(self):
+        print("in diameter getter")
+        return self._diameter
+
+    @diameter.setter
+    def diameter(self, dia):
+        print("in diameter setter")
+        radius_in = dia // 2
+        self.__init__(radius_in)
```
hosungs, line 61: Good job not forgetting this! Yes, on every setter, all the derivative attributes should be recomputed in your design here (which is optimizing for less computation, not recalculating the area, the circumference, and the diameter every time any of those properties is accessed). Otherwise, there'll be inconsistencies among attributes, which you nicely avoided!

students/timm67/session08/circle.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, radius):
+        self._radius = float(radius)
+        self._diameter = float(2.0) * self._radius
+        self._area = pi * (self._radius * self._radius)
+        self._circumference = float(2.0) * pi * self._radius
+
+    @classmethod
+    def from_diameter(self, dia):
+        radius_in = dia // 2
```
hosungs, line 22: Minor: Hmm... Why integer division on `dia`?

students/timm67/session08/circle.py
```diff
@@ -0,0 +1,75 @@
...
+        return Circle(radius_in)
+
+    def __str__(self):
+        return "Circle with radius [{0:6.2f}]".format(self._radius)
+
+    def __repl__(self):
+        return "Circle({0:d})".format(int(self._radius))
```
hosungs, line 29: Why `int(...)`, when `self._radius` can be any float number?

hosungs (commented):
Too bad that there are now 3 assignments all piled up on this PR... It looks like your html_render assignment is yet to be completed (with the HTML tag attributes), so the PR couldn't have been merged/closed earlier. It also looks like you finished the circle assignment quite quickly, and moved on to oo_mailroom right away, all of which are really great to see by the way, but it looks like we'll have to use this PR for all 3 assignments. Just enter this PR's URL in all 3 assignments' Canvas submissions. Thanks!

students/timm67/session07/html_render.py
```diff
@@ -4,28 +4,164 @@
...
+
+    def render(self, out_fd, indent_in=''):
+        try:
+            # render DOCTYPE
+            if isinstance(self, Html) is True:
+                out_fd.write('{0}\n'.format(self._doctype))
+            if(len(self._attributes) > 0):
+                out_fd.write('{0}<{1} '.format(indent_in, self._tag))
+                for key, value in self._attributes.items():
+                    out_fd.write('{0}={1} '.format(key, value))
```
hosungs: You are still missing `"` around `{1}` that'll cause test failures.

students/timm67/session07/html_render.py
```diff
@@ -4,28 +4,164 @@
...
+    def append(self, content_in):
+        raise NotImplementedError
+
+    def render(self, out_fd, indent_in=''):
+        try:
+            if(len(self._attributes) > 0):
+                out_fd.write('{0}<{1} '.format(indent_in, self._tag))
+                for key, value in self._attributes.items():
+                    out_fd.write('{0}={1} '.format(key, value))
+                out_fd.write('> ')
```
hosungs: Not good to have duplicate code, especially when it's buggy (because fixing the same bug on all the duplicate parts is very error-prone). Consider creating a separate function that performs this part only and reuse it as needed.

students/timm67/session07/html_render.py
```diff
@@ -4,28 +4,164 @@
...
-                for line in self.content:
-                    try:
-                        fd.write(line)
-                    except IOError:
-                        print("I/O Error with file [{0}] on writeline".format(out_file))
-                        return
+            if(len(self._attributes) > 0):
+                out_fd.write('{0}<{1} '.format(indent_in, self._tag))
+                for key, value in self._attributes.items():
+                    out_fd.write('{0}={1} '.format(key, value))
```
hosungs: More duplicate code in that same regard...

hosungs (commented):
Looking more, after receiving more commit notification email, but not sure if my comments were addressed, and if any of the open assignments is fully finished... I guess I'll have to leave this PR open, but please continue working on finishing one assignment at a time and let us know. Thanks.

ghost:
@hosungs: I implemented the changes you suggested and that helped me to fix subtle gotchas in the property handling. I have submitted the file(s) for grading. Pull request should be OK to process. Thank you for your help.

students/timm67/session07/html_render.py
```diff
@@ -7,25 +7,201 @@
...
-                        return
+            # render DOCTYPE
+            if isinstance(self, Html) is True:
+                out_fd.write('{0}\n'.format(self._doctype))
+
+            num_attr = self.render_attributes(out_fd, indent_in)
+            if (num_attr > 0):
+                out_fd.write('>\n')
+            else:
+                out_fd.write('\n')
```
hosungs, line 52: The `render_attributes()` is definitely a good change, but I'm not sure why you put these lines (49-62) here, outside `render_attributes()`, when they seem to fit in the `render_attributes()` just fine...?

students/timm67/session07/html_render.py
```diff
@@ -7,25 +7,201 @@
...
+
+    def render_attributes(self, out_fd, indent_in):
+        num_attrib = len(self._attributes)
+        if(num_attrib > 0):
+            out_fd.write('{0}<{1} '.format(indent_in, self._tag))
+            for key, value in self._attributes.items():
+                out_fd.write('{0}="{1}" '.format(key, value))
+        else:
+            out_fd.write('{0}<{1}>'.format(indent_in, self._tag))
+        return num_attrib
```
hosungs: You don't need to (and shouldn't) redefine `render_attributes()` here with the same duplicate code, because this class inherits from `Element`, where `render_attributes()` is already defined with exactly the same code. That's the whole point of inheritance: code reuse and avoidance of duplicate code.

students/timm67/session07/html_render.py
```diff
@@ -7,25 +7,201 @@
...
+
+    def render_attributes(self, out_fd, indent_in):
+        num_attrib = len(self._attributes)
+        if(num_attrib > 0):
+            out_fd.write('{0}<{1} '.format(indent_in, self._tag))
+            for key, value in self._attributes.items():
+                out_fd.write('{0}="{1}" '.format(key, value))
+        else:
+            out_fd.write('{0}<{1} '.format(indent_in, self._tag))
+        return num_attrib
```
hosungs, line 115: Ditto.

hosungs (commented):
The `render_attributes()` is definitely a good change, but there's still unnecessary code duplication. Left some comments on that, and I think it should be addressed for good object-oriented programming practice.

hosungs (commented):
Took a quick look and the `MailroomCli` class looks good! Since this PR couldn't be closed before you tackle all the pending assignments, I'll keep this open until the due is fully over or until you declare you are fully completed. Thanks!

ghost:
Hi @hosungs . Added missing grid printer assignment. I think OO mailroom is in good shape. Both have been submitted. I think it is safe to go ahead with the PR. Thanks for your guidance.

students/timm67/session02/print_grid.py
```diff
@@ -1,10 +1,82 @@
...
+            if (cur_row % column_width) == 0 and (cur_col % column_width) == 0:
+                grid += _intersect
+            elif (cur_row % column_width) == 0:
+                grid += _line
+            elif (cur_col % column_width) == 0:
+                grid += _vertline
+            else:
+                grid += _space
+            cur_col += 1
+        cur_row += 1
```
hosungs, line 64: Why use `cur_row` and `cur_col`, when `i` and `j` seem to be the same corresponding values?

hosungs (commented):
Looked at your `print_grid` submission, and left a comment. It's been quite a long while since it was due, so I may not remember exactly what's expected, but it doesn't look complete yet, based on my cursory memory of the multi-step instructions. Just make sure to submit this to the corresponding Canvas assignment, so that this can be graded/reflected in your final score. Keep pushing new commits as needed, as this PR already contains multiple assignments submissions anyway.

students/timm67/session08/circle.py
ghost: Addressed.

students/timm67/session08/circle.py
ghost: radius input is always an int and converted to float in constructor and subsequent calcs (just wanted to be consistent)

students/timm67/session08/circle.py
ghost: Thanks for the guidance.

students/timm67/session07/html_render.py
ghost: Fixed.

students/timm67/session07/html_render.py
ghost: Fixed

students/timm67/session07/html_render.py
ghost: Minor change I'd like to forgo to avoid code churn. Yes could be in render_attributes()

students/timm67/session07/html_render.py
ghost: dup code issue resolved

students/timm67/session07/html_render.py
ghost: dup code issue resolved

students/timm67/session02/print_grid.py
ghost: Yes, true. Could have used cur_row, and cur_col as iterators. I knew I needed two loops, so I roughed them out quick, and worked on the logic inside the loops. I think I have implemented what was asked for in the assignment and run all the examples in the 'how to' document.

hosungs (commented):
Saw your responses to my comments with more commits. Tried to look them up and verify myself, but it's getting too hard because of so many commits and files already. I'll have to assume that what you addressed are what I intended.

Please make sure to leave a comment in this PR, if you want me to review your future commits and provide feedback. Otherwise, I'll assume those commits are your WIP and wait for your request in a comment. Also, all the grading will be done over Canvas with your final submission, so please do make sure to submit your final versions there in the corresponding Canvas assignment dropboxes. Thanks.

hosungs:
Closing this PR as communicated over email (Tim said he's done).

PR #96, opened 2018-12-04
Session 7 submittals for Dave Arasim

davidarasim:


students/dave_arasim/session07/html_render.py
```diff
@@ -0,0 +1,166 @@
...
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def _open_tag(self):
+        nv_pairs = ''
+
+        if self.attrs is not None:
+            for key,value in self.attrs.items():
+                nv_pairs += ' ' + str(key) + '='
+                nv_pairs += '"' + str(value) + '"'
```
hosungs, line 34: Minor: The whole lines 31-34 could be reduced to just 1 line as follows (using list comprehension):
```
nv_pairs = ''.join(' {}={}'.format(key, value) for key, value in self.attrs.items())
```

students/dave_arasim/session07/html_render.py
```diff
@@ -0,0 +1,166 @@
...
+        return "</{}>".format(self.tag)
+
+    def render(self, out_file):
+        if self.doctype != '':
+            out_file.write(self.doctype)
+        out_file.write(Element.indent_str * Element.indent_level)
+        out_file.write(self._open_tag())
+        out_file.write("\n")
+
+        Element.indent_level += 1
```
hosungs, line 48: This is an interesting approach and possibly works (unless it's used in multiple threads), but this is essentially like using a global variable to keep track of the indentation level. Probably a better/desired approach is to pass the indentation string (a number of blank spaces) in the `render(...)` call, which will call its sub-element's (content's) `render(...)` with an extended indentation string.

hosungs (commented):
Looks good overall, with just a few minor comments. Will merge/close this PR for your continued work. Thanks!

PR #97, opened 2018-12-09
Late assignment for session08 Circle Class

yushusongmsazure:


students/yushus/session08/Circle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __le__(self, other):
+        if isinstance(other, Circle):
+            return self.radius <= other.radius
+        raise TypeError('{other} has to be a Circle type')
+
+    def __eq__(self, other):
+        if isinstance(other, Circle):
+            return self.radius == other.radius
+        raise TypeError('{other} has to be a Circle type')
```
hosungs, line 83: Minor: All the `isinstance(other, Circle)` checks might not be considered very good/Pythonic. I think you can let the caller handle any thrown exception because of passing an incompatible type object, or at least just do `try: ... except:` with recasting the `AttributeError` to something you want (like `TypeError` in your code). No modification is required.

hosungs (approved):
Looks good overall for the circle class assignment, merging/closing this PR. Thanks.

PR #98, opened 2018-12-10
Added Circle assigment

nallarun10:


students/arun_nalla/session08/circle.py
```diff
@@ -0,0 +1,120 @@
...
+""" Circle class assignment by
+Arun Nalla 12/08/2018"""
+
+from math import pi
+import random
+
+class Circle():
+
+    def __init__(self, radius):
+        if radius >= 1:
```
hosungs: Shouldn't this be `if radius > 0:`, if you don't want to allow 0-radius circle? Note that radius doesn't have to be integer. It's a minor point, though, I suppose...

students/arun_nalla/session08/circle.py
```diff
@@ -0,0 +1,120 @@
...
+            self.radius = radius
+        else:
+            raise ValueError ('Value is less than or equal to zero')
+
+    @property
+    def diameter(self):
+        self.diameter = self.radius * 2
+        return self.radius
```
hosungs: Why return `self.radius`, after computing `self.diameter`? Also you don't need to save the computed result to an attribute (`self.diameter`). You can just return the computed value as is (`return self.radius * 2`).

hosungs (commented):
Left a few comments (minor ones that better be addressed). Will wait for them to be addressed, before merging/closing this PR. Thanks.

students/arun_nalla/session08/circle.py
nallarun10: Thanks Housung, added revised file.

hosungs (approved):
Thanks for your revision. Merging/closing this PR as such.

PR #99, opened 2018-12-10
Initial pull request for mailroom_oo

motians:


students/shawn_degraw/session09/mailroom_oo/cli_main.py
```diff
@@ -1,23 +1,66 @@
...
+
+
+def printreport():
+    """ Prints report created in the class """
+
+    print(donor_database.create_report())
+    print()
+
+
+def printthankyou(name, donationamount):
```
hosungs, line 47: Minor: This is OK, but I think this function could/better be a method of the `DonorCollection` class, or even the `Donor` class.

students/shawn_degraw/session09/mailroom_oo/donor_models.py
```diff
@@ -23,18 +22,131 @@ def name(self):
...
+
+    def create_report(self):
+        """ Generates report line by line and returns report as a string """
+
+        reportbody = self.report_header()
+        for donor in self.collect_data():
+            dobject = self.search_name(donor)
+            reportbody = "\n".join([reportbody, "{:<27}${:>11.2f} {:>11d}  ${:>12.2f}".format(dobject.name,
+                                    dobject.total_donations / 100, dobject.number_donations, dobject.average_donation / 100)])
+        return reportbody
```
hosungs: Minor: Better practice to construct a multi-line string output is to use a list:
```
lines = [self.report_header()]
for donr in self.collect_data():
    dobject = self.search_name(donor)
    lines.append("{:<27}${:>11.2f} {:>11d}  ${:>12.2f}".format(dobject.name., dobject.total_donations / 100, dobject.number_donations, dobject.average_donation / 100)
return '\n'.join(lines)
```

hosungs (commented):
Looks good overall, almost ready to get merged/closed, but I'll leave this PR open until the due is over. Thanks!

students/shawn_degraw/session09/mailroom_oo/donor_models.py
motians: Modified create_report to use a list instead of string. Easy fix and still returned a string so no change in my cli_main.py module.

hosungs: Saw it and looks good, though it could have been simpler with a list comprehension. No need to change further, though.

PR #100, opened 2018-12-10
session07 html_render

jjakubiak:
A few questions on the session07 html render assignment before I submit.
1) Are we required to create our own tests or are the pre-created tests in test_html_render sufficient?

2) I get the correct output however there are two tests that fail:

a) def test_title(): fails an assertion test at two lines, assert file_contents.endswith("</title>") and assert "\n" not in file_contents. Why is the title row not allowed to have a line break character? It seems to still provide the correct output.

b) def test_indent(): fails an assertion test at line, assert lines[0].startswith("   <"). The lines list looks correct ['<!DOCTYPE html>', '<html>', '    some content', '</html>'], but it seems that we wouldn't want an indent before the doc type header at line[0]?

Thanks!

### DIFF
--- a/students/jjakub/session07/html_render/html_render.py
+++ b/students/jjakub/session07/html_render/html_render.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+
+"""
+Class: Python 210 B, Au2018
+Exercise: Session 06, Mailroom Part 04
+Student: Jason Jakubiak
+
+A class-based system for rendering html.
+"""
+
+# This is the framework for the base class
+class Element(object):
+    """
+    Abstract base class for rendering an Html element
+    """
+    tag = "html"
+    indent = "    "
+
+    def __init__(self, content=None, **kwargs):
+        self.attributes = kwargs
+        self.contents = []
+        if content is not None:
+            self.contents = [content]
+
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+
+    def attrubutes(self):
+        attrb = "".join([' {}="{}"'.format(key, val) 
+                        for key, val in self.attributes.items()]
+                        )
+        return attrb
+
+
+    def make_open_tag(self, cur_ind=""):
+        open_tag = "<{}{}>".format(self.tag, self.attrubutes())
+        return cur_ind + open_tag
+
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind)
+        out_file.write(self.make_open_tag())
+        out_file.write("\n")
+        for content in self.contents:
+            try:
+                content.render(out_file, cur_ind + self.indent)
+            except AttributeError:
+                out_file.write(cur_ind + self.indent)
+                out_file.write(str(content) + "\n")
+        out_file.write(cur_ind)
+        out_file.write("</{}>\n".format(self.tag))
+
+class OneLineTag(Element):
+    tag = "Title"
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind)
+        out_file.write(self.make_open_tag())
+        for content in self.contents:
+            out_file.write(str(content))
+        out_file.write("</{}>\n".format(self.tag))
+
+class Html(Element):
+    tag = "html"
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write("<!DOCTYPE html>\n")
+        super().render(out_file, cur_ind)
+
+
+class Body(Element):
+    tag = "body"
+
+
+class P(Element):
+    tag = "p"
+
+
+class Head(Element):
+    tag = "head"
+
+
+class Title(OneLineTag):
+    tag = "title"
+
+
+class SelfClosingTag(Element):
+    """
+    Base class for tags without content
+    To render horizontal rule and line breaks
+    """
+
+    def append(self, *args, **kwargs):
+        raise TypeError("Adding content to a self closing tag not allowed")
+
+
+    def render(self, out_file, cur_ind=""):
+        open_tag = self.make_open_tag()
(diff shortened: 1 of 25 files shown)


hosungs:
> A few questions on the session07 html render assignment before I submit.
> 
> 1. Are we required to create our own tests or are the pre-created tests in test_html_render sufficient?

The pre-created tests are sufficient.

> 2. I get the correct output however there are two tests that fail:
> 
> a) def test_title(): fails an assertion test at two lines, assert file_contents.endswith("</title>") and assert "\n" not in file_contents. Why is the title row not allowed to have a line break character? It seems to still provide the correct output.

I usually don't run the code myself (it's not very straightforward to get the unmerged code and run it), but I just read them through, so I may not be able to find out what the issues are, but I'll try to see what's going on.

> 
> b) def test_indent(): fails an assertion test at line, assert lines[0].startswith(" <"). The lines list looks correct ['', '', ' some content', ''], but it seems that we wouldn't want an indent before the doc type header at line[0]?

Same here, but if your `lines` list is like what you gave, then your `lines[0]` is `", "`, which doesn't start with `" <"`, so it looks like the test failure is expected...? Are you sure you think your `lines` list is correct?

> 
> Thanks!

students/jjakub/session07/html_render/html_render.py
```diff
@@ -0,0 +1,139 @@
...
+        if content is not None:
+            self.contents = [content]
+
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+
+    def attrubutes(self):
+        attrb = " ".join([' {}="{}"'.format(key, val)
```
hosungs: Since you already added a leading blank space in `' {}="{}"'`, it'd be better if you just do `.join()` on an empty string (`""`), not a string with one blank space like now (`" "`).

students/jjakub/session07/html_render/test_html_render.py
```diff
@@ -0,0 +1,334 @@
...
+def test_title():
+    t = Title()
+    t.append("This is the title")
+
+    file_contents = render_result(t)
+    print(file_contents)
+    assert("This is the title") in file_contents
+    assert file_contents.startswith("<title>")
+    # assert file_contents.endswith("</title>")
+    # assert "\n" not in file_contents
```
hosungs: Yeah, I'm not sure either why a `<title>...</title>` line is not supposed to end with `\n`. Maybe there's some small technical detail specific to the instruction? Sorry I don't remember that off hand...

students/jjakub/session07/html_render/test_html_render.py
```diff
@@ -0,0 +1,334 @@
...
+    """
+    Tests that the indentation gets passed through to the renderer
+    """
+    html = Html("some content")
+    file_contents = render_result(html, ind="   ").rstrip()  #remove the end newline
+
+    # print(file_contents)
+    lines = file_contents.split("\n")
+    print(file_contents)
+    # assert lines[0].startswith("   <")
```
hosungs: I suppose this is where your other test failure occurs? Could you paste the output of your `print(file_contents)` statement verbatim here?

hosungs (commented):
Left some comments, and have to mention that there are files that are deleted/modified outside your student directory, so this PR needs to be fixed for that before merging/closing it. We should do that together in class this Wed. Thanks.

students/jjakub/session07/html_render/html_render.py
jjakubiak: That makes sense. I've updated the code and will commit with the final version.

students/jjakub/session07/html_render/test_html_render.py
jjakubiak: <!DOCTYPE html>
<html>
    some content
</html>

jjakubiak:
Thanks for the quick reply. This is helpful.

students/jjakub/session07/html_render/html_render.py
hosungs: You still have a blank space in `attrb = " ".join(...)`, I think it should go away (to have only one blank space between attributes).

students/jjakub/session07/html_render/test_html_render.py
hosungs: Shouldn't it be `    <html>some content</html>`?

students/jjakub/session07/html_render/test_html_render.py
```diff
@@ -263,12 +264,13 @@ def test_indent():
     html = Html("some content")
     file_contents = render_result(html, ind="   ").rstrip()  #remove the end newline
 
-    # print(file_contents)
     lines = file_contents.split("\n")
     print(file_contents)
-    # assert lines[0].startswith("   <")
+    # assert lines[0].startswith("   <") 
+        ### Question: The item at line[0] is the docheader, why would there be a indent before that?
```
hosungs: That's because the `file_contents = render_result(html, ind="    ").rstrip()` was made with `ind="    "`, which is to be the base indentation level. I suppose you might have not handled that?

students/jjakub/session07/html_render/test_html_render.py
```diff
@@ -263,12 +264,13 @@ def test_indent():
...
     print(repr(lines[-1]))
     # assert lines[-1].startswith("   <")
+        ### Question: The item at line[-1] is the close html tag, why would there be a indent before that?
```
hosungs: Ditto

hosungs (commented):
Answered the questions in the comments

hosungs (approved):
Good job on html_render and circle class assignments. I assume you are still working on oo_mailroom and I'll close this PR, so please create another PR for your oo_mailroom when you submit it. Thanks.

PR #102, opened 2018-12-12
New pull for maillroom_oo wip and other old homework corrections

bmnguyen:
Please give feedback where you see it.


students/brandon_nguyen/session07/html_render.py
```diff
@@ -23,7 +23,7 @@ def append(self, new_content):
 
     def render(self, out_file, cur_ind=""):
         # loop through the list of contents + recursive render(...)
-        # TODO how to be pure if ELEMENT has abstract_tag instead.
+        # TODO how to be pure if ELEMENT has abstract_tag instead?
```
hosungs, line 26: I missed this comment earlier, but I'm not sure what it means to be "pure"...? What is the `abstract_tag` for, in the first place?

students/brandon_nguyen/session08/Circle.py
```diff
@@ -0,0 +1,54 @@
...
+        self.__radius = r
+        self.__diameter = r*2
+
+    @property
+    def diameter(self):
+        return self.__diameter
+    
+    @diameter.setter
+    def diameter(self, d):
+        self.__radius = d/2
```
hosungs, line 34: You should also update the `self.__diameter` as well. Otherwise, your `self.__radius` and `self.__diameter` might be inconsistent after `c = Circle(5)` and `c.diameter = 4` is executed (for example).

In fact, a better approach might be just to remember radius only, and derive diameter in the property getter, not remembering diameter separately.

hosungs (commented):
Left a few comments. I'll keep this PR open, as it's now containing multiple assignments already and we are at the final week anyway. Please keep pushing your new commits here and I'll review as needed.

students/brandon_nguyen/session09/mailroom_oo/donor_model.py
```diff
@@ -51,15 +68,47 @@ def __str__(self):
...
+    def __init__(self):
+        self.__donor_db = {}
     # We need data structure and other constants for printing
 
-    # method to update donation
-
-    # method to return donor in list getName?
+    @property
+    def donor_db(self):
+        return self.__donor_db
```
hosungs: This (exposing the internal raw data store as is) is not very desirable in object-oriented design. Any needed operations on the internal raw data better be supported through methods, not by exposing it fully like this. Besides, I don't see any place where this `donor_db` property is accessed, so I'm wondering if this is needed in the first place?

students/brandon_nguyen/session09/mailroom_oo/donor_model.py
```diff
@@ -51,15 +68,47 @@ def __str__(self):
...
 
-    # method create report
+    # method to update donation
+    def add_donor_db(self, name, donation=None):
+        """ To add donor into collection"""
+        # How best to handle existing donor?
+        # TODO existing donor only take donation?
```
hosungs: You can check if the passed donor `name` already exists in your `self.__donor_db` or not, using the standard Python `key in dict` operator. Probably better check if the `name` doesn't exist in the DB and then create the entry for the new donor, and then proceed to add donation anyway:
```
if name not in self.__donor_db:
    self.__donor_db[name] = Donor(name)
self.__donor_db[name].add_donation(donation)
```

students/brandon_nguyen/session09/mailroom_oo/donor_model.py
```diff
@@ -51,15 +68,47 @@ def __str__(self):
...
+
+        # assume donation is ok TODO more
+        self.__donor_db[name] = Donor(name)
+        self.__donor_db[name].add_donation(donation)
+
+    def list_donors(self):
+        """ quick way to return a list of names in db """
+        lst_names = []
+        [lst_names.append(name) for name in self.__donor_db.keys()]
```
hosungs: Again, this code pattern works, but not a desired one (showing lack of full understanding of the list comprehension). Please change the two lines above to this:
```
lst_names = [name for name in self.__donor_db.keys()]
```
However, the list comprehension is not even needed at all in this case. You can simply say as follows, to get a list of all keys in a dict:
```
lst_names = list(self.__donor_db.keys())
```
This would be considered better than the ones above.

hosungs (commented):
Left comments after receiving new commits notification.

students/brandon_nguyen/session09/mailroom_oo/donor_model.py
bmnguyen: Thank you!  This is much, much cleaner!

students/brandon_nguyen/session08/Circle.py
bmnguyen: thank you! will do.

students/brandon_nguyen/session09/mailroom_oo/donor_model.py
bmnguyen: I used it in one of my test to learn more.  I suppose I can comment it out.

students/brandon_nguyen/session09/mailroom_oo/donor_model.py
bmnguyen: Thank you.  I corrected that.  Please let me know if you see anything else. I just submit my "final" version.  Knock on wood!

